PR #8, opened 2016-07-29
Added support for Last-Modified and ETag

### LINKED ISSUE #2: Cache Validation
| Q | A |
| --- | --- |
| Bug? | no |
| New Feature? | yes |

See php-http/plugins#57, @dbu wrote: 

> the CachePlugin should handle Last-Modified and Etag headers and send If-None-Match resp. If-Modified-Since headers along if the data is still in the cache. this would allow the server to respond with a 304 Not Modified and we could update the cached response and then return that.

Nyholm:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | yes |
| Deprecations? | no |
| Related tickets | fixes #2 |
| Documentation |  |
| License | MIT |
#### What's in this PR?

We should allow the server to answer 304 which allows us to use a response in the cache. 
The BC break here is that all existing items in the cache will be invalid. Is that an issue? It is a **cache** and not a **storage**. 

I introduced a new config parameter `cache_lifetime` that says how long the item should be in cache without it being used. 

I did also add a few parameters to the array we store in the cache such as `expiresAt`, `createdAt`, `etag`. 
#### To Do
- [x] Fix the tests

### DIFF
--- a/spec/CachePluginSpec.php
+++ b/spec/CachePluginSpec.php
@@ -49,13 +49,13 @@ class CachePluginSpec extends ObjectBehavior
         $item->isHit()->willReturn(false);
         $item->expiresAfter(1060)->willReturn($item)->shouldBeCalled();
 
-        $item->set(Argument::that($this->getCacheItemMatcher([
+        $item->set($this->getCacheItemMatcher([
             'response' => $response->getWrappedObject(),
             'body' => $httpBody,
             'expiresAt' => 0,
             'createdAt' => 0,
             'etag' => []
-        ])))->willReturn($item)->shouldBeCalled();
+        ]))->willReturn($item)->shouldBeCalled();
         $pool->save(Argument::any())->shouldBeCalled();
 
         $next = function (RequestInterface $request) use ($response) {
@@ -115,13 +115,13 @@ class CachePluginSpec extends ObjectBehavior
         $pool->getItem('d20f64acc6e70b6079845f2fe357732929550ae1')->shouldBeCalled()->willReturn($item);
         $item->isHit()->willReturn(false);
 
-        $item->set(Argument::that($this->getCacheItemMatcher([
+        $item->set($this->getCacheItemMatcher([
                 'response' => $response->getWrappedObject(),
                 'body' => $httpBody,
                 'expiresAt' => 0,
                 'createdAt' => 0,
                 'etag' => []
-            ])))->willReturn($item)->shouldBeCalled();
+            ]))->willReturn($item)->shouldBeCalled();
         // 40-15 should be 25 + the default 1000
         $item->expiresAfter(1025)->willReturn($item)->shouldBeCalled();
         $pool->save($item)->shouldBeCalled();
@@ -133,9 +133,17 @@ class CachePluginSpec extends ObjectBehavior
         $this->handleRequest($request, $next, function () {});
     }
 
+
+    /**
+     * Private function to match requests
+     *
+     * @param array $expectedData
+     *
+     * @return \Closure
+     */
     private function getCacheItemMatcher(array $expectedData)
     {
-        return function(array $actualData) use ($expectedData) {
+        return Argument::that(function(array $actualData) use ($expectedData) {
             foreach ($expectedData as $key => $value) {
                 if (!isset($actualData[$key])) {
                     return false;
@@ -151,6 +159,6 @@ class CachePluginSpec extends ObjectBehavior
                 }
             }
             return true;
-        };
+        });
     }
 }
--- a/spec/CachePluginSpec.php
+++ b/spec/CachePluginSpec.php
@@ -15,7 +15,7 @@ class CachePluginSpec extends ObjectBehavior
 {
     function let(CacheItemPoolInterface $pool, StreamFactory $streamFactory)
     {
-        $this->beConstructedWith($pool, $streamFactory, ['default_ttl'=>60]);
+        $this->beConstructedWith($pool, $streamFactory, ['default_ttl'=>60, 'cache_lifetime'=>1000]);
     }
 
     function it_is_initializable(CacheItemPoolInterface $pool)
@@ -41,11 +41,12 @@ class CachePluginSpec extends ObjectBehavior
         $response->getBody()->willReturn($stream);
         $response->getHeader('Cache-Control')->willReturn(array());
         $response->getHeader('Expires')->willReturn(array());
+        $response->getHeader('ETag')->willReturn(array());
 
         $pool->getItem('d20f64acc6e70b6079845f2fe357732929550ae1')->shouldBeCalled()->willReturn($item);
         $item->isHit()->willReturn(false);
-        $item->set(['response' => $response, 'body' => $httpBody])->willReturn($item)->shouldBeCalled();
-        $item->expiresAfter(60)->willReturn($item)->shouldBeCalled();
+        $item->set()->willReturn($item)->shouldBeCalled();
+        $item->expiresAfter(1060)->willReturn($item)->shouldBeCalled();
         $pool->save($item)->shouldBeCalled();
 
         $next = function (RequestInterface $request) use ($response) {
@@ -53,6 +54,12 @@ class CachePluginSpec extends ObjectBehavior
         };
 
         $this->handleRequest($request, $next, function () {});
+        $item->get()->shouldHaveKeyWithValue('response', $response);
+        $item->get()->shouldHaveKeyWithValue('body', $httpBody);
+        $item->get()->shouldHaveKey('expiresAt');
+        $item->get()->shouldHaveKey('createdAt');
+        $item->get()->shouldHaveKey('etag');
+
     }
 
     function it_doesnt_store_failed_responses(CacheItemPoolInterface $pool, CacheItemInterface $item, RequestInterface $request, ResponseInterface $response)
--- a/spec/CachePluginSpec.php
+++ b/spec/CachePluginSpec.php
@@ -16,7 +16,10 @@ class CachePluginSpec extends ObjectBehavior
 {
     function let(CacheItemPoolInterface $pool, StreamFactory $streamFactory)
     {
-        $this->beConstructedWith($pool, $streamFactory, ['default_ttl'=>60, 'cache_lifetime'=>1000]);
+        $this->beConstructedWith($pool, $streamFactory, [
+            'default_ttl' => 60,
+            'cache_lifetime'=>1000
+        ]);
     }
 
     function it_is_initializable(CacheItemPoolInterface $pool)
--- a/spec/CachePluginSpec.php
+++ b/spec/CachePluginSpec.php
@@ -3,7 +3,6 @@
 namespace spec\Http\Client\Common\Plugin;
 
 use Prophecy\Argument;
-use Prophecy\Comparator\Factory as ComparatorFactory;
 use Http\Message\StreamFactory;
 use Http\Promise\FulfilledPromise;
 use PhpSpec\ObjectBehavior;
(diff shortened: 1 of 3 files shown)


src/CachePlugin.php
```diff
@@ -68,15 +68,42 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
...
+
         return $next($request)->then(function (ResponseInterface $response) use ($cacheItem) {
+            if (304 === $response->getStatusCode()) {
+                // The cached response we have is still valid
+                $data = $cacheItem->get();
+                $data['expiresAt'] = time() + $this->getMaxAge($response);
+                $cacheItem->set($data)->expiresAfter($this->config['cache_lifetime'] + $data['expiresAt']);
+                $this->pool->save($cacheItem);
+
+                return $this->createResponseFromCacheItem($cacheItem);
```
sagikazarmark, line 110: What if the server responds with 304 but there is no cache?

src/CachePlugin.php
```diff
@@ -86,8 +113,16 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
...
 
-                $cacheItem->set(['response' => $response, 'body' => $body])
-                    ->expiresAfter($this->getMaxAge($response));
+                $expiresAt = time() + $this->getMaxAge($response);
+                $cacheItem
+                    ->expiresAfter($this->config['cache_lifetime'] + $expiresAt)
+                    ->set([
+                    'response' => $response,
+                    'body' => $body,
+                    'expiresAt' => $expiresAt,
```
sagikazarmark: Why do we need this? The cache includes expiration?

sagikazarmark:
Why do we need cache lifetime? Normally the server tells us how long it should be cached, doesn't it?

src/CachePlugin.php
Nyholm: Yes but that we would like to be able to ask the server if the response is still valid after it has expired. If so, we need to store the response for a longer time that it's max age.

src/CachePlugin.php
Nyholm, line 110: We do now that it is a cache. Or else the `If-Modified-Since` and `If-None-Match` would not have been added. And correct me if I'm wrong but I don't think the server will answer 304 without those headers in the request.

Nyholm, line 110: Yes, those headers may have been added before. I'll just make sure to pass the response then.

Nyholm:
Yes, but we would like to have the possibility to serve an expired response if the server answers 304.

Nyholm:
T0: We make a request. Server responds saying the response is valid until T10. Return response form server
T5: We make a request. The response is in the cache and we know it is valid until T10. Return response from cache
T15: We make a request. The response is in the cache but is stale. We ask the server about the response and server answer 304 and that the response is valid until T25. Return response from cache
T20: We make a request. The response is in the cache and we know it is valid until T25. Return response from cache

spec/CachePluginSpec.php
```diff
@@ -41,18 +41,25 @@ function it_caches_responses(CacheItemPoolInterface $pool, CacheItemInterface $i
         $response->getBody()->willReturn($stream);
         $response->getHeader('Cache-Control')->willReturn(array());
         $response->getHeader('Expires')->willReturn(array());
+        $response->getHeader('ETag')->willReturn(array());
 
         $pool->getItem('e3b717d5883a45ef9493d009741f7c64')->shouldBeCalled()->willReturn($item);
         $item->isHit()->willReturn(false);
-        $item->set(['response' => $response, 'body' => $httpBody])->willReturn($item)->shouldBeCalled();
-        $item->expiresAfter(60)->willReturn($item)->shouldBeCalled();
+        $item->set()->willReturn($item)->shouldBeCalled();
```
Nyholm: I can't figure out a good way to write this line. I cant specify the exact data since it is dependent on `time()`. 

Can I say `$item->set(<any argument>)->willRetu....`?

Nyholm:
I've figured out the tests and I found some bugs. 

Im done with the code changes and this PR is ready for review.

src/CachePlugin.php
```diff
@@ -194,11 +231,73 @@ private function getMaxAge(ResponseInterface $response)
     private function configureOptions(OptionsResolver $resolver)
     {
         $resolver->setDefaults([
+            'cache_lifetime' => 2592000, // 30 days
```
GrahamCampbell: might be clearer to actually have an arithmetic expression here since it's only executed once

src/CachePlugin.php
```diff
@@ -68,15 +68,44 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
...
-            $response = $data['response'];
-            $response = $response->withBody($this->streamFactory->createStream($data['body']));
+            if (isset($data['expiresAt']) && time() < $data['expiresAt']) {
+                // This item is still valid according to previous cache headers
+                return new FulfilledPromise($this->createResponseFromCacheItem($cacheItem));
+            }
+
+            // Add headers to ask the server if this cache is still valid
+            if ($mod = $this->getModifiedAt($cacheItem)) {
+                $mod = new \DateTime('@'.$mod);
```
GrahamCampbell: could we use the immutable version here?

GrahamCampbell: Maybe move this creation inside the `getModifiedAt\ method?

src/CachePlugin.php
```diff
@@ -86,8 +115,16 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
...
 
-                $cacheItem->set(['response' => $response, 'body' => $body])
-                    ->expiresAfter($this->getMaxAge($response));
+                $maxAge = $this->getMaxAge($response);
+                $cacheItem
+                    ->expiresAfter($this->config['cache_lifetime'] + $maxAge)
+                    ->set([
+                    'response' => $response,
+                    'body' => $body,
+                    'expiresAt' => time() + $maxAge,
```
GrahamCampbell: We should probably avoid calling `time()` more than once.

Nyholm:
Thank you @GrahamCampbell. I've update the PR according to your suggestions.

spec/CachePluginSpec.php
```diff
@@ -15,7 +16,7 @@ class CachePluginSpec extends ObjectBehavior
 {
     function let(CacheItemPoolInterface $pool, StreamFactory $streamFactory)
     {
-        $this->beConstructedWith($pool, $streamFactory, ['default_ttl'=>60]);
+        $this->beConstructedWith($pool, $streamFactory, ['default_ttl'=>60, 'cache_lifetime'=>1000]);
```
GrahamCampbell: odd cs?

src/CachePlugin.php
```diff
@@ -195,13 +231,78 @@ private function getMaxAge(ResponseInterface $response)
...
+     * @return string|null
+     */
+    private function getETag(CacheItemInterface $cacheItem)
+    {
+        $data = $cacheItem->get();
+        if (!isset($data['etag'])) {
+            return;
+        }
+
+        if (is_array($data['etag'])) {
```
GrahamCampbell: Maybe negate this and return the bottom thing to avoid the nesting, and also would avoid needing that extra return

Nyholm: Thank you. I did not remove the extra return however. No biggie but `void !== null`.

spec/CachePluginSpec.php
```diff
@@ -16,7 +16,10 @@ class CachePluginSpec extends ObjectBehavior
 {
     function let(CacheItemPoolInterface $pool, StreamFactory $streamFactory)
     {
-        $this->beConstructedWith($pool, $streamFactory, ['default_ttl'=>60, 'cache_lifetime'=>1000]);
+        $this->beConstructedWith($pool, $streamFactory, [
+            'default_ttl' => 60,
+            'cache_lifetime'=>1000
```
GrahamCampbell: missing spaces?

src/CachePlugin.php
```diff
@@ -293,16 +293,16 @@ private function getETag(CacheItemInterface $cacheItem)
...
 
-            return;
+        foreach ($data['etag'] as $etag) {
+            if (!empty($etag)) {
+                return $etag;
+            }
         }
 
-        return $data['etag'];
+        return;
```
GrahamCampbell: don't need this bit now?

GrahamCampbell: it's fine to just fall out of the function

src/CachePlugin.php
```diff
@@ -195,13 +231,78 @@ private function getMaxAge(ResponseInterface $response)
...
+            return $data['etag'];
+        }
+
+        foreach ($data['etag'] as $etag) {
+            if (!empty($etag)) {
+                return $etag;
+            }
+        }
+
+        return;
```
GrahamCampbell: don't need this statement?

src/CachePlugin.php
Nyholm: Yes, it would be the same result if we remove this. But that is technically wrong. Because `void !== null`.

GrahamCampbell: actually, it's fine because php will internally make it null whenever you try to use it, you just can't go the other way around if you wanted void but returned null

src/CachePlugin.php
Nyholm: See https://github.com/php-http/cache-plugin/pull/8#discussion_r73135923

GrahamCampbell: yeh, i replied there

src/CachePlugin.php
Nyholm: Okey, let's remove it then

src/CachePlugin.php
```diff
@@ -87,8 +114,17 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
...
                 }
 
-                $cacheItem->set(['response' => $response, 'body' => $body])
-                    ->expiresAfter($this->getMaxAge($response));
+                $maxAge = $this->getMaxAge($response);
+                $currentTime = time();
+                $cacheItem
+                    ->expiresAfter($this->config['cache_lifetime'] + $maxAge)
+                    ->set([
+                    'response' => $response,
```
GrahamCampbell: is this one level not enough indented?

GrahamCampbell:
Looks good. I've not physically tested this in the real world yet though.

src/CachePlugin.php
Nyholm: Thank you. But shouldn't StyleCi pick this up?

GrahamCampbell: Symfony has no rules about indentation like this I think?

src/CachePlugin.php
```diff
@@ -87,8 +114,17 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
                     $response = $response->withBody($this->streamFactory->createStream($body));
                 }
 
-                $cacheItem->set(['response' => $response, 'body' => $body])
-                    ->expiresAfter($this->getMaxAge($response));
+                $maxAge = $this->getMaxAge($response);
+                $currentTime = time();
+                $cacheItem
+                    ->expiresAfter($this->config['cache_lifetime'] + $maxAge)
```
GrahamCampbell, line 125: Why are we doing this please?

Nyholm, line 125: We would like to serve Responses that the server once said should expire. 

See https://github.com/php-http/cache-plugin/pull/8#issuecomment-236175746

src/CachePlugin.php
```diff
@@ -69,15 +69,42 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
         $cacheItem = $this->pool->getItem($key);
 
         if ($cacheItem->isHit()) {
-            // return cached response
             $data = $cacheItem->get();
-            $response = $data['response'];
-            $response = $response->withBody($this->streamFactory->createStream($data['body']));
+            if (isset($data['expiresAt']) && time() < $data['expiresAt']) {
```
dbu, line 78: is the isset check only for BC with old cache items? we should add a task to remove the isset in version 2 if that is the case.

src/CachePlugin.php
```diff
@@ -69,15 +69,42 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
...
 
-            return new FulfilledPromise($response);
+            if ($etag = $this->getETag($cacheItem)) {
+                $request = $request->withHeader('If-None-Match', $etag);
+            }
         }
 
         return $next($request)->then(function (ResponseInterface $response) use ($cacheItem) {
+            if (304 === $response->getStatusCode()) {
+                if (!$cacheItem->isHit()) {
```
dbu, line 95: would this be because the userland code did an If-None-Match / If-Modified-Since header? but indeed i think this is the only thing we can do at this point.

dbu:
went over the code, looks good to me.

for a moment i was wondering about [stale-while-validate](https://tools.ietf.org/html/rfc5861) but this is just content validation and not advanced proxy operations. in async mode we theoretically could use stale-while-validate to return the expired response and re-fetch to our local cache. but this has nothing to do with this PR and anyways seems like a bad use case - one should use a dedicated caching proxy for such things, not implement this on client side.

src/CachePlugin.php
dbu, line 125: when doing cache validation, we keep outdated caches and send If-Modified-Since / If-None-Match requests to the server to check if the cache is ok to be used. we might add some comment in the code though, to make it clear.

src/CachePlugin.php
Nyholm, line 95: Yes. If the users sends a request with a If-None-Match / If-Modified-Since header we do not have it in cache and we just have to return the response.

src/CachePlugin.php
Nyholm, line 78: Hm. Removing the isset in 2.0 will not help when we fetch a stored v1 response and try to fetch it as v2. 

We could add a cache busting parameter when we create the cache key. 
Or we could make sure to always run isset even on 'response' and 'body'

Nyholm:
TODO: 
- [x] Decide on how to handle updates in the array that is stored in the cache. `isset` or cache busting https://github.com/php-http/cache-plugin/pull/8#discussion_r73156317

src/CachePlugin.php
dbu, line 78: in my opinion, deploying a major version change could have the requirement to clear the cache.

as we add this now, i would assume this to be a non-issue, as cache arrays will be updated by the time we release version 2, unless somebody jumps from 1.0 to 2.0 directly.

Nyholm, line 78: Okey, I'll keep using the issets.

Nyholm:
We use issets. 

This PR is ready to be merged

src/CachePlugin.php
dbu, line 78: can you add a code comment for the isset please? its a bit confusing when not thinking about BC with existing cache items

sagikazarmark:
Thanks, will tag a release soon.

Nyholm:
Awesome. Thank you for all the reviews and feedback.

PR #10, opened 2016-08-01
The max age must be an integer

GrahamCampbell:
| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | no |
| BC breaks? | no |
| Deprecations? | no |
| Related tickets | partially #5 |
| License | MIT |
#### What's in this PR?

This method must not return `null` ever since the psr cache interface does not allow null as a cache age.
#### Why?

After having issues, I created https://github.com/madewithlove/illuminate-psr-cache-bridge/issues/1, then after investigation, I came here, and found https://github.com/php-http/cache-plugin/pull/5 which appears to fix this, but the phpdoc was left unchanged, so needs correcting to avoid confusion or future breaks.
#### Checklist
- [ ] Updated CHANGELOG.md to describe BC breaks / deprecations | new feature | bugfix

### DIFF
--- a/src/CachePlugin.php
+++ b/src/CachePlugin.php
@@ -77,7 +77,7 @@ final class CachePlugin implements Plugin
         }
 
         return $next($request)->then(function (ResponseInterface $response) use ($cacheItem) {
-            if ($this->isCacheable($response)) {
+            if ($this->isCacheable($response) && ($maxAge = $this->getMaxAge($response)) > 0) {
                 $bodyStream = $response->getBody();
                 $body = $bodyStream->__toString();
                 if ($bodyStream->isSeekable()) {
@@ -87,7 +87,7 @@ final class CachePlugin implements Plugin
                 }
 
                 $cacheItem->set(['response' => $response, 'body' => $body])
-                    ->expiresAfter($this->getMaxAge($response));
+                    ->expiresAfter($maxAge);
                 $this->pool->save($cacheItem);
             }
 
--- a/src/CachePlugin.php
+++ b/src/CachePlugin.php
@@ -163,7 +163,7 @@ final class CachePlugin implements Plugin
     private function getMaxAge(ResponseInterface $response)
     {
         if (!$this->config['respect_cache_headers']) {
-            return $this->config['default_ttl'];
+            return (int) $this->config['default_ttl'];
         }
 
         // check for max age in the Cache-Control header
@@ -171,7 +171,7 @@ final class CachePlugin implements Plugin
         if (!is_bool($maxAge)) {
             $ageHeaders = $response->getHeader('Age');
             foreach ($ageHeaders as $age) {
-                return $maxAge - ((int) $age);
+                return ((int) $maxAge) - ((int) $age);
             }
 
             return (int) $maxAge;
@@ -183,7 +183,7 @@ final class CachePlugin implements Plugin
             return (new \DateTime($header))->getTimestamp() - (new \DateTime())->getTimestamp();
         }
 
-        return $this->config['default_ttl'];
+        return (int) $this->config['default_ttl'];
     }
 
     /**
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
My last commit should fully fix this, and fix a potential issue with php 7.1's new number treatement.

GrahamCampbell:
I probably need to add a couple of tests here tbh.

src/CachePlugin.php
```diff
@@ -158,20 +158,20 @@ private function createCacheKey(RequestInterface $request)
...
      * @param ResponseInterface $response
      *
-     * @return int|null
+     * @return int
      */
     private function getMaxAge(ResponseInterface $response)
     {
         if (!$this->config['respect_cache_headers']) {
-            return $this->config['default_ttl'];
+            return (int) $this->config['default_ttl'];
```
sagikazarmark, line 166: Since the default value is null, this will become 0. Is that a good idea? Also, we have OptionsResolver in the plugin. @Nyholm any ide why null is allowed here?

GrahamCampbell, line 166: Yes, `null` needs to become `0`. `null` is now allowed as an expiry time.

GrahamCampbell, line 166: Surely if the age is null, we shouldn't bother caching it anyways?

sagikazarmark, line 166: Good question. Also, I think 0/null usually means unlimited cache ttl?

GrahamCampbell, line 166: Ok, I've updated the implementation. I'll probably turn off caching in my production app right now until this is resolved.

Nyholm:
According to the PSR-6 specification, `null` is a perfectly fine value for the `expiresAfter` method. The doc block says: 

>    /**
>      \* Sets the expiration time for this cache item.
>      *
>      \* @param int|\DateInterval $time
>      \*   The period of time from the present after which the item MUST be considered
>      \*   expired. An integer parameter is understood to be the time in seconds until
>      \*   expiration. If null is passed explicitly, a default value MAY be used.
>      \*   If none is set, the value should be stored permanently or for as long as the
>      \*   implementation allows.
>      *
>      \* @return static
>      \*   The called object.
>      */
>     public function expiresAfter($time);

Setting it to 0 means that there is no caching att all. 

I suggest to close this issue and open an issue on the cache implementation. Btw, I've created a test suite for PSR-6 caches that the implementation may use. http://www.php-cache.com/en/latest/implementing-cache-pools/integration-tests/

EDIT:
Hm... they are using my tests. I guess I need to add a new test case then.

Nyholm:
Tests added: https://github.com/php-cache/integration-tests/pull/59
PR sent to the cache implementation: https://github.com/madewithlove/illuminate-psr-cache-bridge/pull/4

@GrahamCampbell If it is okey with you, please close this issue.

GrahamCampbell:
This is still incorrect. We should never want to cache a request forever.

GrahamCampbell:
Perhaps our default ttl is just bad?

sagikazarmark:
Null ttl does not mean cache forever. It means fall back to default AND cache forever if there is none. At least that's what I can see in the not-so-good PSR-6 documentation. Perhaps @Nyholm can verify?

GrahamCampbell:
Yeh, that's what it sounds like. I don't think we should trust that behaviour, and should set the default ourselves.

Nyholm:
if `respect_cache_headers=false` then you probably want to set the `default_ttl` to something. 

> Null ttl does not mean cache forever. It means fall back to default AND cache forever if there is none. 

Yes, it means fallback to the default. That is up to the implementation to provide a default. At PHP-cache we do not use a default, we store it forever.

sagikazarmark:
I don't think we can choose a "sane" default ttl as it is always context dependent.

Nyholm:
> Yeh, that's what it sounds like. I don't think we should trust that behaviour, and should set the default ourselves.

That is why we have `default_ttl`. It allows the user to specify the default. Do you mean we should have a default value for `default_ttl`?

GrahamCampbell:
I guess not. We're doing nothing wrong here.

Issue #4, opened 2016-05-04
Vary


sagikazarmark:
| Q | A |
| --- | --- |
| Bug? | no |
| New Feature? | yes |

See php-http/plugins#59

Nyholm:
@dbu wrote:

> the CachePlugin currently does not respect the Vary header. vary by its nature can not be part of the cache key (we don't know the vary from the request, only from the response). but we should check the response we find for a Vary header and compare the headers from the original request with the ones from the current request.
> 
> in many scenarios, we will never vary (e.g. different accept-encodings are unlikely) but e.g. when sending requests on behalf of a logged in user, we might do vary.

I agree that we should be more sophisticated when creating the cache key. 

Since we cant put the data from the vary header in the cache, it would mean that we have to make a request to the server an all request.. so vary has no effect. 

A possible solution is to have a storage/repository with cache keys, vary headers and responses. So at each request we look in the storage. If cached vary header exists we validates those headers on the request and return a response if we got one.

dbu:
i think what we need is a cache that would give us all variants we currently have. we would look at each candidate and check our headers for the vary headers and if they match. if we find no match, we send the request and add that variant to the cache. we could probably make the assumption that all variants have to have the same vary headers - then we could move the definition of what this varies on to the cache entry instead of looking at each response. would have to check the rfc about the vary header.

the thing is that with vary, requests will lead to the same cache entry regardless of vary, because at that point we don't know what to vary on. the cache entry then must provide several possible responses, distinguished by the headers as defined in vary.

Nyholm:
Yeah, that was pretty much was I was trying to say.

joelwurtz:
So this will be the following ? : 

ResponseCache -> Cache Response without it's body
BodyCache -> Cache Body of response

ResponseCache key = targetUrl of the request (there can be multiple responses for a same targetUrl, so we store an array of Response)
BodyCache key = vary hashing (we took the headers to vary on from the response, their values from the request, create a cache key and retrieve the body from the cache if exists, or make the actual request if not)

dbu:
we could port (or reuse?) the store that symfony HttpCache uses: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpKernel/HttpCache/Store.php#L140-L165

symfony HttpCache\Store has two flaws for us:
- it is written around the symfony Request and not PSR-7. we could write an adapter to solve this but meh we would convert request and response between psr and symfony.
[...]

PR #12, opened 2016-08-01
Use sha1 to reduce the risk of collisions

GrahamCampbell:
| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | no |
| BC breaks? | no |
| Deprecations? | no |
| License | MIT |
#### What's in this PR?

Switches the key generation algorithm over to use `sha1` rather than `md5`.
#### Why?

The performance is similar, and the collision risk is lower.
#### Example Usage

image
#### Checklist
- [ ] Updated CHANGELOG.md to describe BC breaks / deprecations | new feature | bugfix
#### To Do
- [ ] If the PR is not complete but you want to discuss the approach, list what remains to be done here

### DIFF
--- a/src/CachePlugin.php
+++ b/src/CachePlugin.php
@@ -77,7 +77,7 @@ final class CachePlugin implements Plugin
         }
 
         return $next($request)->then(function (ResponseInterface $response) use ($cacheItem) {
-            if ($this->isCacheable($response)) {
+            if ($this->isCacheable($response) && ($maxAge = $this->getMaxAge($response)) > 0) {
                 $bodyStream = $response->getBody();
                 $body = $bodyStream->__toString();
                 if ($bodyStream->isSeekable()) {
@@ -87,7 +87,7 @@ final class CachePlugin implements Plugin
                 }
 
                 $cacheItem->set(['response' => $response, 'body' => $body])
-                    ->expiresAfter($this->getMaxAge($response));
+                    ->expiresAfter($maxAge);
                 $this->pool->save($cacheItem);
             }
 
--- a/src/CachePlugin.php
+++ b/src/CachePlugin.php
@@ -39,6 +39,7 @@ final class CachePlugin implements Plugin
      *
      *     @var bool $respect_cache_headers Whether to look at the cache directives or ignore them
      *     @var int $default_ttl If we do not respect cache headers or can't calculate a good ttl, use this value.
+     *     @var string $hash_algo The hashing algorithm to use when generating cache keys.
      * }
      */
     public function __construct(CacheItemPoolInterface $pool, StreamFactory $streamFactory, array $config = [])
@@ -150,7 +151,7 @@ final class CachePlugin implements Plugin
      */
     private function createCacheKey(RequestInterface $request)
     {
-        return sha1($request->getMethod().' '.$request->getUri());
+        return hash($this->config['hash_algo'], $request->getMethod().' '.$request->getUri());
     }
 
     /**
@@ -196,9 +197,11 @@ final class CachePlugin implements Plugin
         $resolver->setDefaults([
             'default_ttl' => null,
             'respect_cache_headers' => true,
+            'hash_algo' => 'sha1',
         ]);
 
         $resolver->setAllowedTypes('default_ttl', ['int', 'null']);
         $resolver->setAllowedTypes('respect_cache_headers', 'bool');
+        $resolver->setAllowedValues('hash_algo', hash_algos());
     }
 }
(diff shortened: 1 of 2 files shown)


sagikazarmark:
Maybe we can make this configurable? Valid values are the results of `hash_algos()`.

GrahamCampbell:
Maybe we should just allow people to implement a key generator interface and inject that into the constructor and have it do `createCacheKey` from a message. If you're ok with that, I can make that abstraction now.

sagikazarmark:
Hm, okay, but I would be happier if we could make it kind of optional: add a hash option in config defaulting to sha1, allowed values are null and result of hash_algos. (createCacheKey in null case should throw a LogicException if no hash and no CacheKeyGenerator instances are provided). Also, a custom CacheKeyGenerator (name ok?) that could be provided via a setter (with a HashCacheKeyGenerator implementation used as the default).

IMO in most cases such implementation might be an overkill, so I think it would be better to provide a "fallback"/default solution.

WDYT?

GrahamCampbell:
Why would anyone want to customize this? If they do, a CacheKeyGenerator would be fine?

GrahamCampbell:
PS I really dislike setters. I prefer things to be immutable after instantiation.

sagikazarmark:
> PS I really dislike setters. I prefer things to be immutable after instantiation.

Me too, but considering that this adds an extra configuration effort, this seems to be a half-way solution to me.

Also, changing the constructor would be a BC break (unless the generator is optional?), while adding an extra config option and an optionally used setter is not.

GrahamCampbell:
I was thinking making it optional, with a default value of null, then filling it using a ternary statement, yeh.

GrahamCampbell:
Hmm, just looked at the constructor signature. I think using the options resolver might be better.

sagikazarmark:
> Hmm, just looked at the constructor signature. I think using the options resolver might be better.

Hmm, right. You can still fall back to a string then using the hash implementation?

GrahamCampbell:
I've implemented the hash algo selection. How does that look?

src/CachePlugin.php
```diff
@@ -77,7 +78,7 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
         }
 
         return $next($request)->then(function (ResponseInterface $response) use ($cacheItem) {
-            if ($this->isCacheable($response)) {
+            if ($this->isCacheable($response) && ($maxAge = $this->getMaxAge($response)) > 0) {
```
sagikazarmark: Isn't this from another PR?

GrahamCampbell: Err, oops...

GrahamCampbell: Doing this from the github online editor, lol.

sagikazarmark: 😝

sagikazarmark:
Can you please add change log? We follow keep a change log format.

GrahamCampbell:
Wanted to avoid conflicts with other PRs. Do you mind adding something on merge?

Nyholm:
Thank you. 👍

sagikazarmark:
Looks good to me, thanks @GrahamCampbell

egeloen:
I'm a little bit late here but this PR is a BC break in a minor version. All my caches are become broken due to the new hash algo used for the cache key generation.

@sagikazarmark Can you confirm it has been done by mistake?

Anyway, as the new tag has been released, let it as it is, I just need to change my cache config.

GrahamCampbell:
I don't think it would be considered broken to just loose items from the cache?

dbu:
did you get errors, or just cache misses? i think cache misses on a minor version sound ok. i hope we did not release it as a patch release though, that would be unexpected.

sagikazarmark:
I agree, cache misses doesn't sound like BC break.

egeloen:
Yes, I probaby abuse when saying BC break as technically everything is working fine (just differently)... I use it in a specific context: caching http requests done to a third service secured by IPs and use the generated cache for tests execution on Travis but after upgrading, my cache is becomes broken. As already said, not a big deal to fix, just want to discuss your BC policy.

Nyholm:
That is an interesting question. Our general BC promise is here: http://php-http.readthedocs.io/en/latest/httplug/backwards-compatibility.html

But should we consider the cache keys as a part of the BC promise? Whatever we decide should be documented. 

I suggest that we are allowed to update the cache keys for minor versions but not for patches. But I would try to avoid it. 

For the record: 
This PR also updated the cache keys: #8 (merged and released)
And the PR for this issue will likely update the cache keys: #4

joelwurtz:
I would say the following : 

Do not make our keys part of our BC promise, but consider them as it was the case.
So we can make changes to the key in a minor or patch version, but we should try our best to avoid that.

sagikazarmark:
Actually this specific PR could even be a bugfix: with massive usage it's possible that some requests collide using md5.

PR #16, opened 2016-08-03
changelog for cache validation

dbu:
we missed to add a changlog entry in #8. is this good enough, anything to add?


Nyholm:
Great. Thank you.

PR #20, opened 2016-08-13
Allow a cache item to be stored forever

### LINKED ISSUE #19: Not respect cache headers
Hey!

It seems since the last changes, I'm not able to don't consider headers at all.

Here, my set up:

``` php
$this->client = new PluginClient(new Client(), [
    new ErrorPlugin(),
    new CachePlugin(
        new FilesystemAdapter('', 0, __DIR__.'/cache'),
        new GuzzleStreamFactory(),
        ['respect_cache_headers' => false]
    ),
]);
```

After debuggint it, I would update https://github.com/php-http/cache-plugin/blob/master/src/CachePlugin.php#L78 by:

``` php
if (!$this->config['respect_cache_headers'] || (isset($data['expiresAt']) && time() < $data['expiresAt'])) {
```

WDYT?

Nyholm:
| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | no |
| BC breaks? | no |
| Deprecations? | no |
| Related tickets | fixes #19 |
| Documentation | https://github.com/php-http/documentation/pull/144 |
| License | MIT |
#### What's in this PR?

If you configure your `default_ttl` to be null you should store (and be valid) the response as long as the cache item is stored. 

Also, if you configure `cache_lifetime` to be null you should store the cache item forever. 
#### Why?

At the moment we calculate the `expiresAt` as `$maxAge + time()` which is just `time()` when $maxAge is null. This is why we got issues like #19.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Change Log
 
+## UNRELEASED
+
+### Fixed
+
+- Issue when you use `respect_cache_headers=>false` in combination with `default_ttl=>null`.
 
 ## 1.1.0 - 2016-08-04
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,14 @@
 
 ## UNRELEASED
 
+### Changed
+
+- The default value for ``default_ttl`` is changed from ``null`` to ``0``. 
+
 ### Fixed
 
 - Issue when you use `respect_cache_headers=>false` in combination with `default_ttl=>null`.
+- We allow ``cache_lifetime`` to be set to ``null``.
 
 ## 1.1.0 - 2016-08-04
 
--- a/src/CachePlugin.php
+++ b/src/CachePlugin.php
@@ -159,7 +159,7 @@ final class CachePlugin implements Plugin
      *
      * @param int|null $maxAge
      *
-     * @return int|null Unix system time
+     * @return int|null Unix system time. A null value means that the response expires when the cache item expires.
      */
     private function calculateResponseExpiresAt($maxAge)
     {
--- a/src/CachePlugin.php
+++ b/src/CachePlugin.php
@@ -155,7 +155,7 @@ final class CachePlugin implements Plugin
 
     /**
      * Calculate the timestamp when a response expires. After that timestamp, we need to send a
-     * If-Modified-Since / If-None-Match request to validate the response
+     * If-Modified-Since / If-None-Match request to validate the response.
      *
      * @param int|null $maxAge
      *
--- a/src/CachePlugin.php
+++ b/src/CachePlugin.php
@@ -144,7 +144,7 @@ final class CachePlugin implements Plugin
     private function calculateCacheItemExpiresAfter($maxAge)
     {
         if ($this->config['cache_lifetime'] === null && $maxAge === null) {
-            return null;
+            return;
         }
 
         return $this->config['cache_lifetime'] + $maxAge;
@@ -157,11 +157,11 @@ final class CachePlugin implements Plugin
      */
     private function getResponseExpiresAt($maxAge)
     {
-        if ($maxAge !== null) {
-            return time() + $maxAge;
+        if ($maxAge === null) {
+            return;
         }
 
-        return null;
+        return time() + $maxAge;
     }
 
     /**
--- a/src/CachePlugin.php
+++ b/src/CachePlugin.php
@@ -159,7 +159,7 @@ final class CachePlugin implements Plugin
      *
      * @param int|null $maxAge
      *
-     * @return int|null Unix system time. A null value means that the response expires when the cache item expires.
+     * @return int|null Unix system time. A null value means that the response expires when the cache item expires
      */
     private function calculateResponseExpiresAt($maxAge)
     {
--- a/src/CachePlugin.php
+++ b/src/CachePlugin.php
@@ -74,8 +74,8 @@ final class CachePlugin implements Plugin
 
         if ($cacheItem->isHit()) {
             $data = $cacheItem->get();
-            // The isset() is to be removed in 2.0.
-            if (isset($data['expiresAt']) && time() < $data['expiresAt']) {
+            // The array_key_exists() is to be removed in 2.0.
+            if (array_key_exists('expiresAt', $data) && ($data['expiresAt'] === null || time() < $data['expiresAt'])) {
                 // This item is still valid according to previous cache headers
                 return new FulfilledPromise($this->createResponseFromCacheItem($cacheItem));
             }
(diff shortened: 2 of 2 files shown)


egeloen:
@Nyholm :+1: It fixes #19

dbu:
i think we should change the default_ttl to something like 10 minutes, otherwise we will start caching responses without caching instructions forewever, which could be unexpected.

Nyholm:
> i think we should change the default_ttl to something like 10 minutes,

If we are using `respect_cache_headers=>true` and there is no cache headers. We fallback on the default_ttl. The default behavior should be that we do not cache at all. (Since there is not cache headers) The user has to configure the default_ttl himself. 

I've changed the default value of default_ttl to 0.

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Change Log
 
+## UNRELEASED
+
+### Changed
+
+- The default value for ``default_ttl`` is changed from ``null`` to ``0``.
```
dbu, line 7: as null did not work as expected anyways, i think this change is ok. there only is a problem if somebody explicitly created the plugin with (null, false) - but imo that falls into the category of a bugfix that can be a BC break if you relied on the bug.

dbu:
looks good to me. any objections? otherwise lets squash and merge this.

sagikazarmark:
No objections, but we have to make sure that there are no regressions.

David Buchmann notifications@github.com ezt írta (2016. augusztus 15.,
hétfő):

CHANGELOG.md
Nyholm, line 7: Agreed

dbu:
> No objections, but we have to make sure that there are no regressions.

yes. i think it looks right, did you see any problematic thing? do you
have a suggestion what we should do additionally to prevent regressions?

sagikazarmark:
No, I haven't spotted anything. Generally speaking our last few features
introduced some regressions, which is bad. I would rather maintain BC than
adding new shiny features. Just a note, nothing special for this PR.

David Buchmann notifications@github.com ezt írta (2016. augusztus 15.,
hétfő):

src/CachePlugin.php
```diff
@@ -138,6 +137,34 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
...
+    private function calculateCacheItemExpiresAfter($maxAge)
+    {
+        if ($this->config['cache_lifetime'] === null && $maxAge === null) {
+            return;
+        }
+
+        return $this->config['cache_lifetime'] + $maxAge;
+    }
+
+    /**
```
dbu: lets add phpdoc here. "Calculate the timestamp when a response expires. After that timestamp, we need to send a If-Modified-Since / If-None-Match request to validate the response"

src/CachePlugin.php
```diff
@@ -138,6 +137,34 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
...
+     * @param int|null $maxAge
+     *
+     * @return int|null
```
dbu: Unix system time?

src/CachePlugin.php
```diff
@@ -138,6 +137,34 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
     }
 
     /**
```
dbu: Calculate the timestamp when this cache item should be dropped from the cache. 

This can be higher than the HTTP max-age when using cache validation.

dbu:
looking over the code again, i wonder two things: does it make sense to cache for longer than max-age if we have no information for cache validation (Last-Modified or Etag)? it seems like in that case we can't reuse the cached item anyways. maybe too complicated, and in the future we could be adding support for grace operations. (stale-on-error, stale-while-revalidate)

~~and would it ever make sense to configure default ttl to a value higher than the cache lifetime? probably not, but again not critical.~~ (we calculate the cache lifetime as maxage + cache lifetime, so if anything we err on caching too long)

RE regressions: i think this is a bugfix, not just a new feature, so we should do it. and if we feel not confident about breaking behaviour, we need to check on the test coverage and maybe add more tests.

src/CachePlugin.php
```diff
@@ -138,6 +137,34 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
...
+     * @param int|null $maxAge
+     *
+     * @return int|null
```
dbu: unix system time? `@see PSR for caching`?

src/CachePlugin.php
```diff
@@ -138,6 +137,40 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
     }
 
     /**
+     * Calculate the timestamp when this cache item should be dropped from the cache. The lowest value return after
+     * the calculation is $maxAge.
```
Nyholm: FYI, I did not follow your suggestion here.

Nyholm:
> looking over the code again, i wonder two things: does it make sense to cache for longer than max-age if we have no information for cache validation (Last-Modified or Etag)? it seems like in that case we can't reuse the cached item anyways. maybe too complicated,

You are correct. We "waste" cache here. But I would like to leave that as a possible future improvement. And yes, it could be too complicated. 

> and in the future we could be adding support for grace operations. (stale-on-error, stale-while-revalidate)

Yes. =)

src/CachePlugin.php
```diff
@@ -138,6 +137,40 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
...
+     *
+     * @param int|null $maxAge
+     *
+     * @return int|null Unix system time. @see PSR6 for caching
```
dbu: sorry, `@see` should be on a separate line. maybe make this `@return int|null Unix system time passed to the PSR-6 cache`?

Nyholm: Perfect

src/CachePlugin.php
dbu: "The lowest value that can be returned is $maxAge"

src/CachePlugin.php
dbu: should we mention that null means the cache item never expires and is only evicted if the cache drops items before they expire?

src/CachePlugin.php
```diff
@@ -138,6 +137,40 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
...
+        return $this->config['cache_lifetime'] + $maxAge;
+    }
+
+    /**
+     * Calculate the timestamp when a response expires. After that timestamp, we need to send a
+     * If-Modified-Since / If-None-Match request to validate the response
+     *
+     * @param int|null $maxAge
+     *
+     * @return int|null Unix system time
```
dbu: lets explain that null means it never expires

src/CachePlugin.php
Nyholm: I do not think so because 
- it is a private method, 
- it is described in the docs
- it is not really true

`null` means that the cache implementation store it for its default value. If you are using a memory (array) cache it will be dropped after the request. 

I think it is enough to reference the PSR-6

src/CachePlugin.php
Nyholm: Yes, agreed

src/CachePlugin.php
dbu: then lets replace the whole return to "This value is passed as cache lifetime to the PSR-6 cache". and not even talk about unix system time.

being private does not mean it does not need to be documented. i usually don't remember such things a year later even if i coded it myself, and here contributors might try to understand the code, so good documentation is valuable even on internal things.

Nyholm: Done. 

Sure, private functions should be properly documented. I think one can exclude some edge cases and be a little bit shorter than on public methods. But I guess you are correct. It is a lot of numbers in this class already...

dbu:
thanks a lot @Nyholm !

Nyholm:
Thank you for merging

PR #21, opened 2016-08-15
Bumped branch alias

GrahamCampbell:


dbu:
i think you are right, #20 should have bumped the alias as its not an urgent fix, more of a DX fix and making features work that previously where not useable as intended.

GrahamCampbell:
It's a new feature, so per semver, the next release should be minor, not patch.

dbu:
its an edge case, as one could argue it was meant to work like this but never did work.

Issue #3, opened 2016-05-04
Cache-Control header handling


sagikazarmark:
| Q | A |
| --- | --- |
| Bug? | no |
| New Feature? | yes |

See php-http/plugins#58

GrahamCampbell:
👍

dbu:
we do respect max-age. what is missing is a clearer role concept of how we handle s-maxage / private / no-cache / no-store instructions. as discussed in the other issue, we should probably have two plugins for the roles "single client / browser" and "caching proxy / shared cache". ideas for good names?

Anahkiasen:
Any progress on this issue? Trying to use the plugin in the context of the `php-github-api` package but most (all?) Github responses return the `private` directive which makes the plugin skip caching :/

dbu:
glad if you can work on it. i think the aproach of having two plugins for the two roles (cache proxy / personal cache) seems cleanest

acrobat:
I'm also running into this issue while caching github api responses. So I would like to put some work in to this, but how should we fix this?

Deprecate the general `CachePlugin` and have 2 new cache plugins for "single client" and "shared cache"? Or keep the currect `CachePlugin` and add the 2 extra caching type classes?

And if I understand correctly the only difference between the 2 new caching types will be the handling of private and no-store?

dbu:
i think we mainly should add things to the $options to customize the behaviour. a config field for the list of cache-control directives should be respected, instead of the boolean in respect_cache_headers. for BC, we can rewrite respect_cache_headers=false to an empty list. the default value for respect_cache_headers should become the headers we currently look at.

once we have that, we can do a ClientCachePlugin and ServerCachePlugin that set the right defaults for respect_cache_headers, to make code more explicit. but people can also use CachePlugin directly for special situations. in #24 @tuupola is adding support to configure which request methods can be cached.

@php-http/owners what do you think of this approach?

acrobat:
Sound good to me! Let's see what @php-http/owners think about it, after that I will try to start work on it!

tuupola:
Would also be nice to have support for request cache headers. Currently they are ignored. For example when client sends request with `Cache-Control: no-cache` fresh content should be requested from the server.

sagikazarmark:
The options solution sounds like a good idea to me.

I am not sure about separate plugin classes though. If it's only about special construction, how about just having static constructors? `CachePlugin::clientCache` and `CachePlugin::serverCache`

dbu:
[...]

PR #24, opened 2017-02-10
Optionally enable caching of any request method

tuupola:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Documentation   | https://github.com/php-http/documentation/pull/174
| License         | MIT

#### What's in this PR?

Adds  new `methods` setting which allows you to change which request methods are allowed to be cached.

#### Why?

Sometimes it is desirable to cache `POST` requests. For example when using HTTPlug based SOAP client. I am currently using this in live system for SOAP requests.

#### Example Usage

To enable caching of `GET` and `POST` requests you can do the following.

```php
$cache = new CachePlugin($pool, $streamFactory, [
    "default_ttl" => 3600 * 48,
    "respect_cache_headers" => true,
    "methods" => ["GET", "POST"]
]);
```

#### Checklist

- [x] Updated CHANGELOG.md to describe new feature.
- [x] Documentation pull request created.

#### To Do

- [x] Should PUT and PATCH et al. also be allowed if user requests so?
- [x] Make sure tests are actually ok.

### DIFF
--- a/spec/CachePluginSpec.php
+++ b/spec/CachePluginSpec.php
@@ -97,8 +97,14 @@ class CachePluginSpec extends ObjectBehavior
         $this->handleRequest($request, $next, function () {});
     }
 
-    function it_stores_post_requests_when_allowed(CacheItemPoolInterface $pool, CacheItemInterface $item, RequestInterface $request, ResponseInterface $response, StreamFactory $streamFactory, StreamInterface $stream)
-    {
+    function it_stores_post_requests_when_allowed(
+        CacheItemPoolInterface $pool,
+        CacheItemInterface $item,
+        RequestInterface $request,
+        ResponseInterface $response,
+        StreamFactory $streamFactory,
+        StreamInterface $stream
+    ) {
         $this->beConstructedWith($pool, $streamFactory, [
             'default_ttl' => 60,
             'cache_lifetime' => 1000,
@@ -116,9 +122,9 @@ class CachePluginSpec extends ObjectBehavior
 
         $response->getStatusCode()->willReturn(200);
         $response->getBody()->willReturn($stream);
-        $response->getHeader('Cache-Control')->willReturn(array())->shouldBeCalled();
-        $response->getHeader('Expires')->willReturn(array())->shouldBeCalled();
-        $response->getHeader('ETag')->willReturn(array())->shouldBeCalled();
+        $response->getHeader('Cache-Control')->willReturn([])->shouldBeCalled();
+        $response->getHeader('Expires')->willReturn([])->shouldBeCalled();
+        $response->getHeader('ETag')->willReturn([])->shouldBeCalled();
 
         $pool->getItem('e4311a9af932c603b400a54efab21b6d7dea7a90')->shouldBeCalled()->willReturn($item);
         $item->isHit()->willReturn(false);
--- a/spec/CachePluginSpec.php
+++ b/spec/CachePluginSpec.php
@@ -166,6 +166,9 @@ class CachePluginSpec extends ObjectBehavior
         $this
             ->shouldThrow("Symfony\Component\OptionsResolver\Exception\InvalidOptionsException")
             ->during('__construct', [$pool, $streamFactory, ['methods' => ['GET', 'HEAD"', 'POST']]]);
+        $this
+            ->shouldThrow("Symfony\Component\OptionsResolver\Exception\InvalidOptionsException")
+            ->during('__construct', [$pool, $streamFactory, ['methods' => ['GET', 'head', 'POST']]]);
     }
 
     function it_calculate_age_from_response(CacheItemPoolInterface $pool, CacheItemInterface $item, RequestInterface $request, ResponseInterface $response, StreamInterface $stream)
--- a/spec/CachePluginSpec.php
+++ b/spec/CachePluginSpec.php
@@ -49,7 +49,7 @@ class CachePluginSpec extends ObjectBehavior
         $response->getHeader('Expires')->willReturn(array())->shouldBeCalled();
         $response->getHeader('ETag')->willReturn(array())->shouldBeCalled();
 
-        $pool->getItem('15a7c0e71475fda62536abd940c55a9d1749ce47')->shouldBeCalled()->willReturn($item);
+        $pool->getItem('d20f64acc6e70b6079845f2fe357732929550ae1')->shouldBeCalled()->willReturn($item);
         $item->isHit()->willReturn(false);
         $item->expiresAfter(1060)->willReturn($item)->shouldBeCalled();
 
@@ -79,7 +79,7 @@ class CachePluginSpec extends ObjectBehavior
         $response->getHeader('Cache-Control')->willReturn(array());
         $response->getHeader('Expires')->willReturn(array());
 
-        $pool->getItem('15a7c0e71475fda62536abd940c55a9d1749ce47')->shouldBeCalled()->willReturn($item);
+        $pool->getItem('d20f64acc6e70b6079845f2fe357732929550ae1')->shouldBeCalled()->willReturn($item);
         $item->isHit()->willReturn(false);
 
         $next = function (RequestInterface $request) use ($response) {
@@ -188,7 +188,7 @@ class CachePluginSpec extends ObjectBehavior
         $response->getHeader('Expires')->willReturn(array());
         $response->getHeader('ETag')->willReturn(array());
 
-        $pool->getItem('15a7c0e71475fda62536abd940c55a9d1749ce47')->shouldBeCalled()->willReturn($item);
+        $pool->getItem('d20f64acc6e70b6079845f2fe357732929550ae1')->shouldBeCalled()->willReturn($item);
         $item->isHit()->willReturn(false);
 
         $item->set($this->getCacheItemMatcher([
@@ -225,7 +225,7 @@ class CachePluginSpec extends ObjectBehavior
         $response->getHeader('Expires')->willReturn(array());
         $response->getHeader('ETag')->willReturn(array('foo_etag'));
 
-        $pool->getItem('15a7c0e71475fda62536abd940c55a9d1749ce47')->shouldBeCalled()->willReturn($item);
+        $pool->getItem('d20f64acc6e70b6079845f2fe357732929550ae1')->shouldBeCalled()->willReturn($item);
         $item->isHit()->willReturn(false);
         $item->expiresAfter(1060)->willReturn($item);
 
@@ -258,7 +258,7 @@ class CachePluginSpec extends ObjectBehavior
 
         $response->getStatusCode()->willReturn(304);
 
-        $pool->getItem('15a7c0e71475fda62536abd940c55a9d1749ce47')->shouldBeCalled()->willReturn($item);
+        $pool->getItem('d20f64acc6e70b6079845f2fe357732929550ae1')->shouldBeCalled()->willReturn($item);
         $item->isHit()->willReturn(true, false);
         $item->get()->willReturn([
             'response' => $response,
@@ -283,7 +283,7 @@ class CachePluginSpec extends ObjectBehavior
         $request->getUri()->willReturn('/');
         $request->getBody()->shouldBeCalled();
 
-        $pool->getItem('15a7c0e71475fda62536abd940c55a9d1749ce47')->shouldBeCalled()->willReturn($item);
+        $pool->getItem('d20f64acc6e70b6079845f2fe357732929550ae1')->shouldBeCalled()->willReturn($item);
         $item->isHit()->willReturn(true);
         $item->get()->willReturn([
             'response' => $response,
@@ -322,7 +322,7 @@ class CachePluginSpec extends ObjectBehavior
         // Make sure we add back the body
         $response->withBody($stream)->willReturn($response)->shouldBeCalled();
 
-        $pool->getItem('15a7c0e71475fda62536abd940c55a9d1749ce47')->shouldBeCalled()->willReturn($item);
+        $pool->getItem('d20f64acc6e70b6079845f2fe357732929550ae1')->shouldBeCalled()->willReturn($item);
         $item->isHit()->willReturn(true, true);
         $item->expiresAfter(1060)->willReturn($item)->shouldBeCalled();
         $item->get()->willReturn([
--- a/spec/CachePluginSpec.php
+++ b/spec/CachePluginSpec.php
@@ -50,7 +50,7 @@ class CachePluginSpec extends ObjectBehavior
         $response->getHeader('Expires')->willReturn(array())->shouldBeCalled();
         $response->getHeader('ETag')->willReturn(array())->shouldBeCalled();
 
-        $pool->getItem('d20f64acc6e70b6079845f2fe357732929550ae1')->shouldBeCalled()->willReturn($item);
+        $pool->getItem('15a7c0e71475fda62536abd940c55a9d1749ce47')->shouldBeCalled()->willReturn($item);
         $item->isHit()->willReturn(false);
         $item->expiresAfter(1060)->willReturn($item)->shouldBeCalled();
 
@@ -80,7 +80,7 @@ class CachePluginSpec extends ObjectBehavior
         $response->getHeader('Cache-Control')->willReturn(array());
         $response->getHeader('Expires')->willReturn(array());
 
-        $pool->getItem('d20f64acc6e70b6079845f2fe357732929550ae1')->shouldBeCalled()->willReturn($item);
+        $pool->getItem('15a7c0e71475fda62536abd940c55a9d1749ce47')->shouldBeCalled()->willReturn($item);
         $item->isHit()->willReturn(false);
 
         $next = function (RequestInterface $request) use ($response) {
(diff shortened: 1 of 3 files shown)


src/CachePlugin.php
```diff
@@ -273,12 +276,17 @@ private function configureOptions(OptionsResolver $resolver)
...
+            'methods' => ['GET', 'HEAD'],
         ]);
 
         $resolver->setAllowedTypes('cache_lifetime', ['int', 'null']);
         $resolver->setAllowedTypes('default_ttl', ['int', 'null']);
         $resolver->setAllowedTypes('respect_cache_headers', 'bool');
+        $resolver->setAllowedTypes('methods', 'array');
         $resolver->setAllowedValues('hash_algo', hash_algos());
+        $resolver->setAllowedValues('methods', function ($value) {
+            return 0 === count(array_diff($value, ['GET', 'HEAD', 'POST']));
```
dbu: i think once we allow to configure, we should allow any methods. we need a big warning sign for this anyways, as people should know what they do before caching post and other potentially modifying requests... for example, it could make sense to cache a PROPFIND request (webdav, its a non-modifying request)...

src/CachePlugin.php
```diff
@@ -225,6 +224,10 @@ private function getCacheControlDirective(ResponseInterface $response, $name)
      */
     private function createCacheKey(RequestInterface $request)
     {
+        if ('POST' === $request->getMethod()) {
+            return hash($this->config['hash_algo'], $request->getMethod().' '.$request->getUri().' '.$request->getBody());
```
dbu: i don't like that we specifically do this for POST (as i want to support configuring all request methods as cacheable). i think using the body in hash should not be an issue - its usually just empty. technically, GET requests can have a body as well, for example with elasticsearch. so its actually an improvement to also use the body for the hash.

dbu (approved):
my first reaction was "wtf, why?", then i read soap and think thats a reasonable usecase. and after some more thinking, i actually feel we should allow arbitrary methods to be cached if we allow to configure that at all.

dbu:
btw, caching proxies like varnish usually do a GET when seeing a HEAD request they have no cache for. then they store the response and return just the headers to honor the HEAD request. but i don't think we want to do that here. if people will do a GET anyways, they should not do a HEAD request first. they might be implementing their own cache, but then they should probably not use the cache-plugin.

src/CachePlugin.php
tuupola: I agree. Question is since RFC2616 says valid request method can be pretty much any string (not sure if A..Z or alphanumeric) should any request method be allowed or just the common methods defined in [RFC2616-sec9](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html)

src/CachePlugin.php
tuupola: Agree on this one too. However some might consider changing the way how keys are generated a BC break. See discussion in #12. Although if I read it correctly the consensus in the end was this would not be considered a BC break.

dbu: i would not consider that a BC break. chances are you replace the cache on a new deployment anyways, and even if not, its just a cache.

src/CachePlugin.php
dbu: i'd not do any validation apart from requiring an array of strings.

spec/CachePluginSpec.php
```diff
@@ -97,6 +97,49 @@ function it_doesnt_store_post_requests(CacheItemPoolInterface $pool, CacheItemIn
         $this->handleRequest($request, $next, function () {});
     }
 
+    function it_stores_post_requests_when_allowed(CacheItemPoolInterface $pool, CacheItemInterface $item, RequestInterface $request, ResponseInterface $response, StreamFactory $streamFactory, StreamInterface $stream)
```
sagikazarmark: Can you please format this according to PSR-1/2?

spec/CachePluginSpec.php
```diff
@@ -97,6 +97,49 @@ function it_doesnt_store_post_requests(CacheItemPoolInterface $pool, CacheItemIn
...
+        $stream->isSeekable()->willReturn(true);
+        $stream->rewind()->shouldBeCalled();
+
+        $request->getMethod()->willReturn('POST');
+        $request->getUri()->willReturn('/post');
+        $request->getBody()->willReturn($stream);
+
+        $response->getStatusCode()->willReturn(200);
+        $response->getBody()->willReturn($stream);
+        $response->getHeader('Cache-Control')->willReturn(array())->shouldBeCalled();
```
sagikazarmark: Please use short array syntax (StyleCI is turned off for spec files because of the snake_case).

spec/CachePluginSpec.php
```diff
@@ -97,6 +97,49 @@ function it_doesnt_store_post_requests(CacheItemPoolInterface $pool, CacheItemIn
...
+        $response->getHeader('Expires')->willReturn(array())->shouldBeCalled();
+        $response->getHeader('ETag')->willReturn(array())->shouldBeCalled();
+
+        $pool->getItem('e4311a9af932c603b400a54efab21b6d7dea7a90')->shouldBeCalled()->willReturn($item);
```
sagikazarmark: Please split this too

src/CachePlugin.php
sagikazarmark: So what's the verdict? Do we change this to allow any methods?

sagikazarmark (changes requested):

sagikazarmark:
Yeah, this one was a shocker for me at first too. 😄 

For the record: we should probably rethink our BDD(-misuse) strategy and rather use PHPUnit or something. Tests are almost unreadable with all the PSR-7 mocks.

spec/CachePluginSpec.php
tuupola: I was following the CS of the rest of the file. Long `array()` syntax is used elsewhere the spec file.

spec/CachePluginSpec.php
tuupola: There were some other PSR-2 related problems too so I made a separate PR #25.

src/CachePlugin.php
tuupola: Found the spec. RFC2616 says request method should be a token and [RFC7230](https://tools.ietf.org/html/rfc7230#section-3.2.6) defines token to be any visible `USASCII` character except delimiters `"(),/:;<=>?@[\]{}`.

tuupola:
Updated the PR. It is now possible to enable caching for any arbitrary request method which conforms RFC7230. Also request body is always used when creating the cache key.

src/CachePlugin.php
```diff
@@ -273,12 +272,19 @@ private function configureOptions(OptionsResolver $resolver)
...
+            'methods' => ['GET', 'HEAD'],
         ]);
 
         $resolver->setAllowedTypes('cache_lifetime', ['int', 'null']);
         $resolver->setAllowedTypes('default_ttl', ['int', 'null']);
         $resolver->setAllowedTypes('respect_cache_headers', 'bool');
+        $resolver->setAllowedTypes('methods', 'array');
         $resolver->setAllowedValues('hash_algo', hash_algos());
+        $resolver->setAllowedValues('methods', function ($value) {
+            $matches = preg_grep('/[^[:alnum:]!#$%&\'*\/+\-.^_`|~]/', $value);
```
dbu: can you please add a comment here with the RFC number that defines this regular expression? just for future reference and to make it clear we did not come up with a random expression here.

dbu:
cool, i like this. 

commits need to be squashed and we should have a documentation pull request (no big thing, just mention the feature in the right place). then its ready to merge imo

src/CachePlugin.php
dbu: i am horrible with regular expressions. is this forcing upper case? we should either force that or uppercase ourselves when processing the options. PSR-7 forces method to upper case. when you specify something lower case, it would never match.

the doc should explain that if you specify methods, you also need to specify GET if you still want those cached. and that you should do specific clients with different setup if you do soap and other stuff, to be sure only the right requests get cached...

tuupola: It does not force uppercase. `[:alnum:]` is alias for `[0-9A-Za-z]`. [RFC2616](https://tools.ietf.org/html/rfc2616#section-5.1.1) says method is case-insensitive so technically both `get` and `GET` should be valid but just different methods.

I have no strong feelings whether we should force uppercase or not. I think most of the time people probably use uppercase methods.

The regexp basically says is request method has any other characters than these the method is invalid.

tuupola:
Documentation added. GitHub now has squash before merge feature which I think is better than doing squash + force push myself.

dbu:
> It does not force uppercase. |[:alnum:]| is alias for |[0-9A-Za-z]|.
> RFC2616 <https://tools.ietf.org/html/rfc2616#section-5.1.1> says method
> is case-insensitive so technically both |get| and |GET| should be valid
> but just different methods.
> 
> I have no strong feelings whether we should force uppercase or not. I
> think most of the time people probably use uppercase methods.

PSR-7 says: 

     * While HTTP method names are typically all uppercase characters, HTTP
     * method names are case-sensitive and thus implementations SHOULD NOT
     * modify the given string.

but for example guzzle psr-7 says:
https://github.com/guzzle/psr7/blob/5f17a79615bd36d52d3b3ce44075aec02605d8bf/src/Request.php#L92-L97

    public function withMethod($method)
    {
        $new = clone $this;
        $new->method = strtoupper($method);
        return $new;
    }

i am surprised by this, but feel like then we should not force upper case but just make sure all our examples use all caps and maybe mention that the option is case sensitive.

dbu (changes requested):
please change the CachePlugin constructor phpdoc to mention the new option for `method` (and say that its case sensitive)

src/CachePlugin.php
```diff
@@ -45,6 +45,7 @@
      *     @var int $cache_lifetime (seconds) To support serving a previous stale response when the server answers 304
      *              we have to store the cache for a longer time than the server originally says it is valid for.
      *              We store a cache item for $cache_lifetime + max age of the response.
+     *     @var array $methods case insensitive list of request methods which can be cached.
```
dbu: its case sensitive, not insensitive

tuupola: Ah yes of course. Brain malfunction...

dbu (approved):

dbu:
looks good to me.

what do the others think? can we merge?

src/CachePlugin.php
```diff
@@ -225,7 +225,7 @@ private function getCacheControlDirective(ResponseInterface $response, $name)
      */
     private function createCacheKey(RequestInterface $request)
     {
-        return hash($this->config['hash_algo'], $request->getMethod().' '.$request->getUri());
+        return hash($this->config['hash_algo'], $request->getMethod().' '.$request->getUri().$request->getBody());
```
Nyholm: What if the body is super large? Do we care?

Also, I suggest a space (or any other non-common delimiter) between uri and body.

src/CachePlugin.php
```diff
@@ -45,6 +45,7 @@
      *     @var int $cache_lifetime (seconds) To support serving a previous stale response when the server answers 304
      *              we have to store the cache for a longer time than the server originally says it is valid for.
      *              We store a cache item for $cache_lifetime + max age of the response.
+     *     @var array $methods case sensitive list of request methods which can be cached.
```
Nyholm: hm.. Yes, they are case sensitive but lower case will always fail..

`@var array $methods list of request methods which can be cached. Methods should always be upper case.`

src/CachePlugin.php
```diff
@@ -273,12 +273,20 @@ private function configureOptions(OptionsResolver $resolver)
...
         ]);
 
         $resolver->setAllowedTypes('cache_lifetime', ['int', 'null']);
         $resolver->setAllowedTypes('default_ttl', ['int', 'null']);
         $resolver->setAllowedTypes('respect_cache_headers', 'bool');
+        $resolver->setAllowedTypes('methods', 'array');
         $resolver->setAllowedValues('hash_algo', hash_algos());
+        $resolver->setAllowedValues('methods', function ($value) {
+            /* Any VCHAR, except delimiters. RFC7230 sections 3.1.1 and 3.2.6 */
+            $matches = preg_grep('/[^[:alnum:]!#$%&\'*\/+\-.^_`|~]/', $value);
```
Nyholm: This looks really complicated. Why not use `[A-Z]+`. That will also make sure you use upper values. Or am I too specific?

Nyholm (commented):
Great PR. Thanks. I had some questions.

src/CachePlugin.php
tuupola: Without space this is backwards compatible with previous versions if body is empty. Adding a space there will invalidate old caches, but it is of course ok with me if needed.

src/CachePlugin.php
tuupola: Unless I have misunderstood the specs request method can also contain lowercase letters. Valid request methods are described in [RFC7230-3.1.1](https://tools.ietf.org/html/rfc7230#section-3.1.1) which states it is a token. Token is defined in [RFC7230-3.2.6](https://tools.ietf.org/html/rfc7230#section-3.2.6) which says any visible `USASCII` character except delimiters `"(),/:;<=>?@[\]{}`.

src/CachePlugin.php
tuupola: Why it will fail?

src/CachePlugin.php
Nyholm: Great. I did not think about that. How about doing something like: 

```php
$body = (string) $request->getBody()
if (!empty($body)) {
  $body = ' ' . $body;
}

return hash($this->config['hash_algo'], $request->getMethod().' '.$request->getUri().$body);
```

src/CachePlugin.php
Nyholm: Sorry, yes you are correct. (Reading the code and RFC more carefully...). I was wrong assuming that `!#$%&` etc should not be a part of the HTTP method. 

HTTP methods are generally uppercase but PSR-7 specifies that case should not change. I think we are correct by running `strtoupper()`. This implementation will, however, always fail when using a lower case method in the configuration. I suggest changing the regex to fail directly if a lower-case method are used. 

```php
$matches = preg_grep('/[^[A-Z0-9]!#$%&\'*\/+\-.^_`|~]/', $value);
```

src/CachePlugin.php
dbu: when you use the guzzle psr implementation, the method is forced to uppercase. but i pasted the fragment from psr-7 somewhere here that says that the getMethod of the request SHOULD preserve case. so depending on implementations, you could need lower case headers. i think its correct to allow lower case and not transform case.

Nyholm: > i think its correct to allow lower case and not transform case.

But that would lead to unexpected behavior. I think we should do a case insensitive comparison or force input to be uppercase.

tuupola: Cache plugin also [converts the request method to uppercase](https://github.com/php-http/cache-plugin/blob/master/src/CachePlugin.php#L65). While this is not technically correct it is probably ok for 99% of the use cases. I think the question is should we follow RFC or not.

src/CachePlugin.php
dbu: i think we do not need to worry about keeping caches valid over deployments. but nyholms idea is good. as GET usually has empty body, it happens to keep old caches.

regarding big body: this method is only executed on methods that should be cached, right? so body will be usually empty, except for things like soap which hopefully won't be big file uploads. you would not want to cache a file upload, anyways.

src/CachePlugin.php
dbu: were do we strtoupper? i think we should leave the method unchanged, given how psr-7 is specified. as mentioned above, guzzle psr-7 request does not follow the SHOULD of the spec.

src/CachePlugin.php
Nyholm: GET and HEAD has (EDIT: generally) no bodies. If you change the default behavior, then it is on you if you get performance issues.

tuupola: I can add it. Why is the space needed BTW?

src/CachePlugin.php
dbu: oh, missed that. ok, that would be a BC break if we stop doing this...

whatever, lets uppercase the method and force configuration to use uppercase. (thats better than magically uppercasing and confusing people that expected some specific behaviour). i think this will be correct in 99.999999% of the cases. would be surprised if anybody ever wants `GET` to be cached but not `get`...

Nyholm:
See https://github.com/php-http/cache-plugin/pull/24#discussion_r101534179 and the method doc block. I think we should do a case insensitive comparison.

src/CachePlugin.php
Nyholm: Agreed.

src/CachePlugin.php
dbu: GET can have body, for example when talking to elasticsearch: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html

Nyholm: To make a difference between these two requests

```http
POST /foo
Host: www.example.com

bar
```

```http
POST /foob
Host: www.example.com

ar
```

tuupola: I feel we could unconditionally add the space. It is only cache so it is not that big deal if old cache gets invalidated when plugin is updated. What do you think?

dbu:
fine for me

src/CachePlugin.php
Nyholm: Hm, sorry for the late response. 

I would prefer not invalidating cache when we can.. Im using Google Translate paid services and I know I will have to pay 100USD if my cache gets invalidated. 

Technically, this is okey. But we could be more nice to our users. =)

dbu:
> I would prefer not invalidating cache when we can.. Im using Google Translate paid services and I know I will have to pay 100USD if my cache gets invalidated.

> Technically, this is okey. But we could be more nice to our users. =)

ok, i guess that is a good argument in favor of not concating ' '.$body if the body is empty (which it usually is for GET). sorry, can you please adjust that @tuupola ? i think then we are really ready. @sagikazarmark or do you see any issues that need to be adressed?

tuupola:
Done. I will update the docs when I know this is frozen.

dbu:
thanks a lot! lets wrap up the doc PR as well. work for more flexibility will continue in #3

Nyholm:
Thank you @tuupola

PR #25, opened 2017-02-11
PSR-2 compliance

tuupola:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | mentioned in #24
| License         | MIT

#### What's in this PR?

Makes code PSR-2 compliant with the exception of relaxing the `PSR1.Methods.CamelCapsMethodName` for spec files. Additionally uses short array syntax in tests as discussed in #24. Lastly adds PHPCS to vendor and enables running it via `$ composer phpcs`.

I know this is three things in one PR but I feel these are closely related and the last one is easy remove if it is unwanted.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,8 @@
     },
     "require-dev": {
         "phpspec/phpspec": "^2.5",
-        "henrikbjorn/phpspec-code-coverage" : "^1.0"
+        "henrikbjorn/phpspec-code-coverage" : "^1.0",
+        "squizlabs/php_codesniffer": "^2.8"
     },
     "autoload": {
         "psr-4": {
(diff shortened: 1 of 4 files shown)


sagikazarmark:
Thanks for the contribution, although I have a few problems with it:

- such changes should be discussed on the boilerplate repo first
- the spec files are kind of specially treated: they follow the PHPSpec style first, then PSR-1/2: so short array syntax, line breaking apply, camel case, method visibility for example do not.

tuupola:
Ok. I though the camel case rule was the only one which should be relaxed.

PR #26, opened 2017-02-22
Implement caching types

### LINKED ISSUE #3: Cache-Control header handling

acrobat:
| Q               | A
| --------------- | ---
| Bug fix?        | yes and no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #3 
| Documentation   | https://github.com/php-http/documentation/issues/178
| License         | MIT

This is a first simple version to get the discussion going and see what changes are needed here. 

#### What's in this PR?

As discussed in #3 this PR introduces 2 factory methods and an option to specify which cache control directives should be checked

#### Checklist

- [x] Updated CHANGELOG.md to describe BC| new feature / bugfix
- [x] Documentation pull request created https://github.com/php-http/documentation/issues/178
- [x] Add tests

### DIFF
--- a/src/CachePlugin.php
+++ b/src/CachePlugin.php
@@ -9,6 +9,7 @@ use Psr\Cache\CacheItemInterface;
 use Psr\Cache\CacheItemPoolInterface;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
+use Symfony\Component\OptionsResolver\Options;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 
 /**
@@ -34,6 +35,13 @@ final class CachePlugin implements Plugin
     private $config;
 
     /**
+     * Cache directives indicating if a response can not be cached.
+     *
+     * @var array
+     */
+    private $noCacheFlags = ['no-cache', 'private', 'no-store'];
+
+    /**
      * @param CacheItemPoolInterface $pool
      * @param StreamFactory          $streamFactory
      * @param array                  $config        {
@@ -45,7 +53,8 @@ final class CachePlugin implements Plugin
      *     @var int $cache_lifetime (seconds) To support serving a previous stale response when the server answers 304
      *              we have to store the cache for a longer time than the server originally says it is valid for.
      *              We store a cache item for $cache_lifetime + max age of the response.
-     *     @var array $methods list of request methods which can be cached.
+     *     @var array $methods list of request methods which can be cached
+     *     @var array $respect_response_cache_directives list of cache directives this plugin will respect while caching responses.
      * }
      */
     public function __construct(CacheItemPoolInterface $pool, StreamFactory $streamFactory, array $config = [])
@@ -53,9 +62,55 @@ final class CachePlugin implements Plugin
         $this->pool = $pool;
         $this->streamFactory = $streamFactory;
 
+        if (isset($config['respect_cache_headers']) && isset($config['respect_response_cache_directives'])) {
+            throw new \InvalidArgumentException(
+                'You can\'t provide config option "respect_cache_headers" and "respect_response_cache_directives". '.
+                'Use "respect_response_cache_directives" instead.'
+            );
+        }
+
         $optionsResolver = new OptionsResolver();
         $this->configureOptions($optionsResolver);
         $this->config = $optionsResolver->resolve($config);
+    }
+
+    /**
+     * This method will setup the cachePlugin in client cache mode. When using the client cache mode the plugin will
+     * cache responses with `private` cache directive.
+     *
+     * @param CacheItemPoolInterface $pool
+     * @param StreamFactory          $streamFactory
+     * @param array                  $config        For all possible config options see the constructor docs
+     *
+     * @return CachePlugin
+     */
+    public static function clientCache(CacheItemPoolInterface $pool, StreamFactory $streamFactory, array $config = [])
+    {
+        // Allow caching of private requests
+        if (isset($config['respect_response_cache_directives'])) {
+            $config['respect_response_cache_directives'][] = 'no-cache';
+            $config['respect_response_cache_directives'][] = 'max-age';
+            $config['respect_response_cache_directives'] = array_unique($config['respect_response_cache_directives']);
+        } else {
+            $config['respect_response_cache_directives'] = ['no-cache', 'max-age'];
+        }
+
+        return new self($pool, $streamFactory, $config);
+    }
+
+    /**
+     * This method will setup the cachePlugin in server cache mode. This is the default caching behavior it refuses to
+     * cache responses with the `private`or `no-cache` directives.
+     *
+     * @param CacheItemPoolInterface $pool
+     * @param StreamFactory          $streamFactory
+     * @param array                  $config        For all possible config options see the constructor docs
+     *
+     * @return CachePlugin
+     */
+    public static function serverCache(CacheItemPoolInterface $pool, StreamFactory $streamFactory, array $config = [])
+    {
+        return new self($pool, $streamFactory, $config);
     }
 
     /**
@@ -183,11 +238,12 @@ final class CachePlugin implements Plugin
         if (!in_array($response->getStatusCode(), [200, 203, 300, 301, 302, 404, 410])) {
             return false;
         }
-        if (!$this->config['respect_cache_headers']) {
-            return true;
-        }
-        if ($this->getCacheControlDirective($response, 'no-store') || $this->getCacheControlDirective($response, 'private')) {
-            return false;
+
+        $nocacheDirectives = array_intersect($this->config['respect_response_cache_directives'], $this->noCacheFlags);
+        foreach ($nocacheDirectives as $nocacheDirective) {
+            if ($this->getCacheControlDirective($response, $nocacheDirective)) {
+                return false;
+            }
         }
 
         return true;
@@ -242,7 +298,7 @@ final class CachePlugin implements Plugin
      */
     private function getMaxAge(ResponseInterface $response)
     {
-        if (!$this->config['respect_cache_headers']) {
+        if (!in_array('max-age', $this->config['respect_response_cache_directives'], true)) {
             return $this->config['default_ttl'];
         }
 
@@ -276,9 +332,11 @@ final class CachePlugin implements Plugin
         $resolver->setDefaults([
             'cache_lifetime' => 86400 * 30, // 30 days
             'default_ttl' => 0,
+            //Deprecated as of v1.3, to be removed in v2.0. Use respect_response_cache_directives instead
             'respect_cache_headers' => true,
             'hash_algo' => 'sha1',
             'methods' => ['GET', 'HEAD'],
+            'respect_response_cache_directives' => ['no-cache', 'private', 'max-age', 'no-store'],
         ]);
 
         $resolver->setAllowedTypes('cache_lifetime', ['int', 'null']);
@@ -291,6 +349,22 @@ final class CachePlugin implements Plugin
             $matches = preg_grep('/[^A-Z0-9!#$%&\'*\/+\-.^_`|~]/', $value);
 
             return empty($matches);
+        });
+
+        $resolver->setNormalizer('respect_cache_headers', function (Options $options, $value) {
+            if (null !== $value) {
+                @trigger_error('The option "respect_cache_headers" is deprecated since version 1.3 and will be removed in 2.0. Use "respect_response_cache_directives" instead.', E_USER_DEPRECATED);
+            }
+
+            return $value;
+        });
+
+        $resolver->setNormalizer('respect_response_cache_directives', function (Options $options, $value) {
+            if (false === $options['respect_cache_headers']) {
+                return [];
+            }
+
+            return $value;
         });
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,17 @@
 ### Added
 
 - New `methods` setting which allows to configure the request methods which can be cached.
+- New `respect_response_cache_directives` config setting to define specific cache directives to respect when handling responses.
+- Introduced `CachePlugin::clientCache` and `CachePlugin::serverCache` factory methods to easily setup the plugin with 
+the correct config settigns for each usecase.
+
+### Changed
+
+- The `no-cache` directive is now respected by the plugin and will not cache the response
+
+### Deprecated
+
+- The `respect_cache_headers` option is deprecated and will be removed in 2.0. This option is replaced by the new `respect_response_cache_directives` option.
 
 ## 1.2.0 - 2016-08-16
 
--- a/spec/CachePluginSpec.php
+++ b/spec/CachePluginSpec.php
@@ -351,6 +351,54 @@ class CachePluginSpec extends ObjectBehavior
         $this->handleRequest($request, $next, function () {});
     }
 
+    function it_caches_private_responses_when_allowed(
+        CacheItemPoolInterface $pool,
+        CacheItemInterface $item,
+        RequestInterface $request,
+        ResponseInterface $response,
+        StreamFactory $streamFactory,
+        StreamInterface $stream
+    ) {
+        $this->beConstructedThrough('clientCache', [$pool, $streamFactory, [
+            'default_ttl' => 60,
+            'cache_lifetime' => 1000,
+        ]]);
+
+        $httpBody = 'body';
+        $stream->__toString()->willReturn($httpBody);
+        $stream->isSeekable()->willReturn(true);
+        $stream->rewind()->shouldBeCalled();
+
+        $request->getMethod()->willReturn('GET');
+        $request->getUri()->willReturn('/');
+        $request->getBody()->shouldBeCalled();
+
+        $response->getStatusCode()->willReturn(200);
+        $response->getBody()->willReturn($stream);
+        $response->getHeader('Cache-Control')->willReturn(['private'])->shouldBeCalled();
+        $response->getHeader('Expires')->willReturn(array())->shouldBeCalled();
+        $response->getHeader('ETag')->willReturn(array())->shouldBeCalled();
+
+        $pool->getItem('d20f64acc6e70b6079845f2fe357732929550ae1')->shouldBeCalled()->willReturn($item);
+        $item->isHit()->willReturn(false);
+        $item->expiresAfter(1060)->willReturn($item)->shouldBeCalled();
+
+        $item->set($this->getCacheItemMatcher([
+            'response' => $response->getWrappedObject(),
+            'body' => $httpBody,
+            'expiresAt' => 0,
+            'createdAt' => 0,
+            'etag' => []
+        ]))->willReturn($item)->shouldBeCalled();
+        $pool->save(Argument::any())->shouldBeCalled();
+
+        $next = function (RequestInterface $request) use ($response) {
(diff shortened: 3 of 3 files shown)


src/CachePlugin.php
```diff
@@ -58,6 +58,19 @@ public function __construct(CacheItemPoolInterface $pool, StreamFactory $streamF
         $this->config = $optionsResolver->resolve($config);
     }
 
+    public static function clientCache(CacheItemPoolInterface $pool, StreamFactory $streamFactory, array $config = [])
+    {
+        // Allow caching of private requests
+        $config['excluded_directives'] = ['no-cache'];
```
dbu: we should check if the option already exists and add it to the existing list

dbu (commented):
i find the name `excluded_directives` confusing. we seem to not exclude those directives, but use them.

i would call the option `respect_response_cache_directives` and deprecate `respect_cache_headers`.

src/CachePlugin.php
tuupola: Or maybe just allow `respect_cache_headers` to be either boolean or array. If boolean true then assume it is all directives. If false then assume it is empty array.

dbu: the things like no-cache are not headers, but directives. and some can be both on request and/or response, and we should treat that separately to allow the user full control. maybe they want no-cache on request to be respected, but on the response they still want to cache...

acrobat:
@dbu thanks for the feedback! I've updated the code!

src/CachePlugin.php
acrobat: I'm also more in favor of separate options. Because I would avoid mixing option types bool/array

acrobat:
I've tested my patch on a project using the github api (which sends the private cache directive with each request). And I can confirm it works when you construct the cache plugin with the clientCache method.

The phpspec tests fail after the patch, but I will look into this fail later tonight!

src/CachePlugin.php
```diff
@@ -242,7 +268,7 @@ private function createCacheKey(RequestInterface $request)
      */
     private function getMaxAge(ResponseInterface $response)
     {
-        if (!$this->config['respect_cache_headers']) {
+        if (!$this->config['respect_cache_headers'] || !in_array('max-age', $this->config['respect_response_cache_directives'], true)) {
```
dbu: i would do the cleanup (create an empty array in respect_response_cache_directives) of this in the constructor and issue a deprecation warning if respect_cache_headers is defined. if both are defined in the constructor, throw an exception as its wrong to do that.

src/CachePlugin.php
```diff
@@ -186,8 +209,11 @@ protected function isCacheable(ResponseInterface $response)
         if (!$this->config['respect_cache_headers']) {
             return true;
         }
-        if ($this->getCacheControlDirective($response, 'no-store') || $this->getCacheControlDirective($response, 'private')) {
-            return false;
+
+        foreach ($this->config['respect_response_cache_directives'] as $cacheDirective) {
+            if ($this->getCacheControlDirective($response, $cacheDirective)) {
```
dbu: this means that if max-age is set to be respected and defined, we do not cache. i think you need a constant for NO_CACHE_FLAGS and array intersect that with respect_response_cache_directives

src/CachePlugin.php
```diff
@@ -58,6 +59,28 @@ public function __construct(CacheItemPoolInterface $pool, StreamFactory $streamF
         $this->config = $optionsResolver->resolve($config);
     }
 
+    public static function clientCache(CacheItemPoolInterface $pool, StreamFactory $streamFactory, array $config = [])
```
dbu: lets add a phpdoc here that says this sets up the cache plugin in a client mode. it will cache even with cache-control private or no-store.

src/CachePlugin.php
```diff
@@ -276,9 +302,11 @@ private function configureOptions(OptionsResolver $resolver)
         $resolver->setDefaults([
             'cache_lifetime' => 86400 * 30, // 30 days
             'default_ttl' => 0,
+            //Deprecated as of v1.3, to be removed in v2.0. Use respect_response_cache_directives instead
             'respect_cache_headers' => true,
             'hash_algo' => 'sha1',
             'methods' => ['GET', 'HEAD'],
+            'respect_response_cache_directives' => ['no-cache', 'private', 'max-age'],
```
dbu: to stay BC, this should include no-store

dbu (changes requested):
thanks, i think we move in the right direction with this. this will end up being quite flexible, while still being reasonably understandable.

src/CachePlugin.php
acrobat: Yes you are right, I will look into a way to do this. We can't use a constant with an array value as the cacheplugin supports php >=5.4

dbu: ah, right. then private property is fine too

acrobat:
@dbu I've incorporated the feedback and I've added a phpspec test for this new option/behavior

src/CachePlugin.php
```diff
@@ -34,31 +35,83 @@
     private $config;
 
     /**
+     * Cache directives indicating if a response can be cached.
```
Nyholm: "can be cached"? "noCacheFlags"?

You mean cannot be cached?

src/CachePlugin.php
```diff
@@ -34,31 +35,83 @@
...
+     *     @var array $respect_response_cache_directives list of cache directives this plugin will respect while caching responses.
      * }
      */
     public function __construct(CacheItemPoolInterface $pool, StreamFactory $streamFactory, array $config = [])
     {
         $this->pool = $pool;
         $this->streamFactory = $streamFactory;
 
+        if (isset($config['respect_cache_headers']) && $config['respect_response_cache_directives']) {
+            throw new \InvalidArgumentException('You can\'t provide config option "respect_cache_headers" and "respect_response_cache_directives". Use "respect_response_cache_directives" instead.');
```
Nyholm: Generally. Make lines shorter than 100 chars. (PSR-2)

src/CachePlugin.php
```diff
@@ -34,31 +35,83 @@
...
+        }
+
         $optionsResolver = new OptionsResolver();
         $this->configureOptions($optionsResolver);
         $this->config = $optionsResolver->resolve($config);
     }
 
     /**
+     * This method will setup the cachePlugin in client cache mode. When using the client cache mode the plugin will cache responses with `private` cache directive.
```
Nyholm: Same here.. Long lines.

src/CachePlugin.php
```diff
@@ -34,31 +35,83 @@
...
+     *
+     * @param CacheItemPoolInterface $pool
+     * @param StreamFactory          $streamFactory
+     * @param array                  $config
```
Nyholm: "See config for constructor" or something similar.

src/CachePlugin.php
```diff
@@ -34,31 +35,83 @@
...
+            }
+        } else {
+            $config['respect_response_cache_directives'] = ['no-cache', 'max-age'];
+        }
+
+        return new self($pool, $streamFactory, $config);
+    }
+
+    /**
+     * This method will setup the cachePlugin in server cache mode. This is the default caching behavior (refuses to cache responses with the `private`or `no-cache` directives.
```
Nyholm: Missing space

src/CachePlugin.php
```diff
@@ -34,31 +35,83 @@
...
+     * @param StreamFactory          $streamFactory
+     * @param array                  $config
+     *
+     * @return CachePlugin
+     */
+    public static function clientCache(CacheItemPoolInterface $pool, StreamFactory $streamFactory, array $config = [])
+    {
+        // Allow caching of private requests
+        if (isset($config['respect_response_cache_directives'])) {
+            if (!in_array('no-cache', $config['respect_response_cache_directives'], true)) {
```
Nyholm: Correct me if Im wrong but one could simplify this by: 

```php
$config['respect_response_cache_directives'][] = 'no-cache';
$config['respect_response_cache_directives'][] = 'max-age';
array_unique($config['respect_response_cache_directives']);
```
Right?

src/CachePlugin.php
```diff
@@ -34,31 +35,83 @@
...
+                $config['respect_response_cache_directives'][] = ['no-cache'];
```
Nyholm: Should you add an array here? `['no-cache']` vs `'no-cache'`.

src/CachePlugin.php
```diff
@@ -34,31 +35,83 @@
...
+
+        return new self($pool, $streamFactory, $config);
+    }
+
+    /**
+     * This method will setup the cachePlugin in server cache mode. This is the default caching behavior (refuses to cache responses with the `private`or `no-cache` directives.
+     *
+     * @param CacheItemPoolInterface $pool
+     * @param StreamFactory          $streamFactory
+     * @param array                  $config
```
Nyholm: Same here, document what parameters or refer to constructor

src/CachePlugin.php
```diff
@@ -183,11 +236,12 @@ protected function isCacheable(ResponseInterface $response)
         if (!in_array($response->getStatusCode(), [200, 203, 300, 301, 302, 404, 410])) {
             return false;
         }
-        if (!$this->config['respect_cache_headers']) {
```
Nyholm, line 186: This cannot be removed. That will break BC.

src/CachePlugin.php
```diff
@@ -183,11 +236,12 @@ protected function isCacheable(ResponseInterface $response)
...
-            return true;
-        }
-        if ($this->getCacheControlDirective($response, 'no-store') || $this->getCacheControlDirective($response, 'private')) {
```
Nyholm, line 189: Removing this will also break BC, right?

Can we do something smart in the constructor? If `respect_cache_headers` is true then populate `respect_response_cache_directives` with "no-store", "private".

src/CachePlugin.php
```diff
@@ -183,11 +236,12 @@ protected function isCacheable(ResponseInterface $response)
...
-            return false;
+
+        $cacheableDirectives = $this->extractDirectives($this->config['respect_response_cache_directives'], $this->noCacheFlags);
```
Nyholm: $cacheableDirectives can only be one of these four (no matter what). Is this really the intention?

```php
 ['no-cache', 'private']
 ['private']
 ['no-cache']
[]
```

Nyholm (changes requested):
Thank you for this PR. I really like the idea of these factory methods. 

I had some comments on some things I did not understand.

src/CachePlugin.php
acrobat: The directive list contains other values like `max-age, no-store` and possibly other values in the future. So I've added this method to easily extract cache directives if they are provided in the config

src/CachePlugin.php
acrobat: I don't think this breaks BC as the default options include the headers

src/CachePlugin.php
acrobat: On line 361 I've added some BC logic. If this option is false we add an empty array so all the checks are skipped and original behavior is kept

acrobat:
Thanks @Nyholm for the review! I've fixed your comments

src/CachePlugin.php
```diff
@@ -34,31 +35,86 @@
...
+     *
+     * @return CachePlugin
+     */
+    public static function clientCache(CacheItemPoolInterface $pool, StreamFactory $streamFactory, array $config = [])
+    {
+        // Allow caching of private requests
+        if (isset($config['respect_response_cache_directives'])) {
+            $config['respect_response_cache_directives'][] = 'no-cache';
+            $config['respect_response_cache_directives'][] = 'max-age';
+            array_unique($config['respect_response_cache_directives']);
```
dbu: afaik array_unique does not modify the input array. you need to re-assign the value to the field.

src/CachePlugin.php
dbu: i agree with the approach. but i would directly use array_intersect here. as you assign it to a variable, its clear enough without the method. i would call the variable $nocacheDirectives.

src/CachePlugin.php
dbu: agree with acrobat, this is handled below with the directives

src/CachePlugin.php
acrobat: yes you're right! Fixed!

acrobat:
@dbu Fixed your new comments!

src/CachePlugin.php
```diff
@@ -34,31 +35,86 @@
...
+     * @var bool   $respect_cache_headers Whether to look at the cache directives or ignore them. This option is
+     *             deprecated, use `respect_response_cache_directives` instead
+     * @var int    $default_ttl (seconds) If we do not respect cache headers or can't calculate a good ttl, use this
+     *             value
+     * @var string $hash_algo The hashing algorithm to use when generating cache keys
+     * @var int    $cache_lifetime (seconds) To support serving a previous stale response when the server answers 304
+     *             we have to store the cache for a longer time than the server originally says it is valid for.
+     *             We store a cache item for $cache_lifetime + max age of the response.
+     * @var array  $methods list of request methods which can be cached
+     * @var array  $respect_response_cache_directives list of cache directives this plugin will respect while caching responses.
```
dbu: this was indented on purpose, because its the list of fields in the $config array.

src/CachePlugin.php
```diff
@@ -34,31 +35,86 @@
...
+     *             }.
```
dbu: and this was left becaues it ends the list of options

src/CachePlugin.php
```diff
@@ -34,31 +35,86 @@
     private $config;
 
     /**
+     * Cache directives indicating if a response can not be cached.
+     *
+     * @var array
+     */
+    private $noCacheFlags = ['no-cache', 'private'];
```
dbu: i think this list should also have no-store

src/CachePlugin.php
```diff
@@ -276,9 +333,11 @@ private function configureOptions(OptionsResolver $resolver)
         $resolver->setDefaults([
             'cache_lifetime' => 86400 * 30, // 30 days
             'default_ttl' => 0,
+            //Deprecated as of v1.3, to be removed in v2.0. Use respect_response_cache_directives instead
             'respect_cache_headers' => true,
             'hash_algo' => 'sha1',
             'methods' => ['GET', 'HEAD'],
+            'respect_response_cache_directives' => ['no-cache', 'private', 'max-age', 'no-store'],
```
dbu, line 339: looks like adding no-cache is a change over current behaviour. i think we can consider that a bugfix. but please add a note to the changelog so people have a chance of seeing the change.

dbu (changes requested):
we are almost there i think. no-cache was not handled previously, but i consider that a bugfix rather than a BC break.

src/CachePlugin.php
tuupola: AFAIK you can actually cache the result if response has `no-cache` directive. It just means you must always revalidate before actually using the cached result. If origin server returns `304 Not Modified` you can use the cached result.

That said the easy implementation is just not to cache. Only difference is some bandwith might be wasted, but implementation is much easier.

https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.9.1

dbu: right. lets open an issue to implement the full no-cache behaviour in the future. but rather than completely ignoring, completely not caching is at least correct according to the protocol.

acrobat: Ok, I've added the no-store directive and left the no-cache for now

src/CachePlugin.php
acrobat: Yeah I think this was fix for styleCi but indeed I've reverted the indent changes!

src/CachePlugin.php
acrobat: See comment above

src/CachePlugin.php
acrobat: I've added a line about this change in the changelog!

acrobat:
I've did some changes according to the feedback and updated the changelog. @dbu & @Nyholm do you guys have time for a last review as I think we are almost ready here?

src/CachePlugin.php
```diff
@@ -46,19 +54,66 @@
...
      *     @var array $methods list of request methods which can be cached.
+     *     @var array $respect_response_cache_directives list of cache directives this plugin will respect while caching responses.
      * }
      */
     public function __construct(CacheItemPoolInterface $pool, StreamFactory $streamFactory, array $config = [])
     {
         $this->pool = $pool;
         $this->streamFactory = $streamFactory;
 
+        if (isset($config['respect_cache_headers']) && $config['respect_response_cache_directives']) {
```
dbu: the second check also needs to have an isset. as is, this will be a fatal error if respect_response_cache_directives is not set. (and throw no exception if respect_cache_headers is set and the other is `[]`)

dbu (changes requested):
thanks! almost there. i noticed one glitch in the constructor sanity check.

src/CachePlugin.php
acrobat: good catch! Fixed!

acrobat:
Fixed comment from @dbu and squashed/rebase the PR

dbu (approved):
looks good to me now. what do the other maintainers think?

src/CachePlugin.php
Nyholm: Good

Nyholm (approved):
Nice. I really like this PR now =)

Thank you @acrobat

dbu:
styleci is a bit weird. i guess we should remove that dot (i would be ok with it, but styleci should be green and i guess we can't easily make it allow just that one place)

acrobat:
@dbu Yeah it was indeed a bit weird that's why I left it there. But removed the dot and now is styleci passing again!

dbu:
thanks a lot! created https://github.com/php-http/documentation/issues/178 about the documentation. needs not a lot, but would be great to have it mentioned in the doc

acrobat:
Yes indeed, that was still a todo for after the PR was merged. I will try to do that tonight!

src/CachePlugin.php
dbu: created #27 for the full no-cache support

acrobat:
Is there an ETA for the 1.3.0 release? As I would like to implement this feature in KnpLabs/php-github-api

Nyholm:
There is no ETA yet. But I guess it will be sooner rather than later. 

Feel free to start a PR on KnpLabs/php-github-api and I'll be happy to review that one. I'll make sure both libraries release a new tag with this feature.

acrobat:
Thanks for the info @Nyholm! I will start a PR to change the cachePlugin usage

acrobat:
@sagikazarmark Do you maybe have some time to create a new release? Would be awesome to use this feature and the method caching feature from #24! Thanks!

tuupola:
Wait a bit with a release. I just noticed that the regexp in #24 allowes `/` character which should not be there. I will make a bugfix PR tomorrow.

acrobat:
ping @tuupola any time to fix this so we can have a release anytime soon?

sagikazarmark:
> Do you maybe have some time to create a new release?

Sorry totally missed this one. My mailbox is quite full lately.

@tuupola ping me if you are done or need anything from me.

tuupola:
Sorry for the delay. I will do it today.

PR #32, opened 2017-03-23
Prepare for 1.3

Nyholm:


Nyholm:
Ping @tuupola is it okey to tag 1.3 now?

dbu:
could we wrap up #29 possibly? as its a behaviour change, even if its the bugfixing type i would feel more at ease if thats a new minor version than a patch version.

Nyholm:
Yes, lets wait for #29. @theofidry will fix that this weekend. 

@tuupola also had something we wanted to do. See https://github.com/php-http/cache-plugin/pull/26#issuecomment-285100517

PR #33, opened 2017-03-23
Do not allow slash in method name

tuupola:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

#### What's in this PR?

Removes `/` character from regexp which is not allowed by the spec. It was bad copy paster from the RFC.


Nyholm (approved):
Great!

Nyholm:
Just make sure the tests and style CI are happy.

Nyholm:
Thank you

PR #35, opened 2017-03-31
Added CacheKeyGenerator

### LINKED ISSUE #30: Customize cache key generation
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | yes
| Version      | 1.2

#### Actual Behavior

The cache key is currently generated based on the following details:

- hash algo
- request method
- URI
- request body

#### Expected Behavior

One should be able to customize this behaviour.

Example: Separate cache entry per authorization header. 

#### Possible Solutions

Introduce a `CacheKeyGenerator` interface which would receive the request as a parameter.
It should be an optional parameter of the plugin, falling back to the current implementation.

Nyholm:
Introducing CacheKeyGenerator interface. This will make sure one can control how a cache key is generated.

This will fix #30 

```php
// Default will use RequestLineAndBodyGenerator
$cachePlugin = new CachePlugin($pool, $streamFactory);
```

```php
$config = [
  'cache_key_generator' => new MyCustomKeyGenerator(), 
];

$cachePlugin = new CachePlugin($pool, $streamFactory, $config);
```

### DIFF
--- a/spec/CachePluginSpec.php
+++ b/spec/CachePluginSpec.php
@@ -2,6 +2,7 @@
 
 namespace spec\Http\Client\Common\Plugin;
 
+use Http\Client\Common\Plugin\Generator\RequestLineAndBodyGenerator;
 use Prophecy\Argument;
 use Http\Message\StreamFactory;
 use Http\Promise\FulfilledPromise;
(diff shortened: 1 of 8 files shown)


Nyholm:
Im done with this PR now. It is ready for review. 

PS: The commit "Support PSR-4" was a typo. I ment to write "Support PHP 5.4"

src/CachePlugin.php
```diff
@@ -338,12 +342,14 @@ private function configureOptions(OptionsResolver $resolver)
             'hash_algo' => 'sha1',
             'methods' => ['GET', 'HEAD'],
             'respect_response_cache_directives' => ['no-cache', 'private', 'max-age', 'no-store'],
+            'cache_key_generator' => null,
```
Nyholm, line 345: Not sure if I should use null here or if I should instanciate the default generator. Using null will make sure we do not instanceiate unneeded objects.

src/CachePlugin.php
```diff
@@ -282,12 +285,13 @@ private function getCacheControlDirective(ResponseInterface $response, $name)
      */
     private function createCacheKey(RequestInterface $request)
     {
-        $body = (string) $request->getBody();
-        if (!empty($body)) {
-            $body = ' '.$body;
+        if (null === $this->config['cache_key_generator']) {
```
sagikazarmark: I would move this check right after the option resolution. I don't see any benefit in timing here, this will be resolved sooner or later.

src/Generator/RequestLineAndBodyGenerator.php
```diff
@@ -0,0 +1,23 @@
...
+namespace Http\Client\Common\Plugin\Generator;
+
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * Generate a cache key from the request method, URI and body.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class RequestLineAndBodyGenerator implements CacheKeyGenerator
```
sagikazarmark: Can we find a shorter name for this?

spec/CachePluginSpec.php
```diff
@@ -2,6 +2,7 @@
 
 namespace spec\Http\Client\Common\Plugin;
 
+use Http\Client\Common\Plugin\Generator\RequestLineAndBodyGenerator;
```
sagikazarmark: Hm, maybe put this under `...\Plugin\Cache\...`

sagikazarmark (changes requested):
I had a few small comments, but looks ok. My fear is that we are putting too much logic here. Do you think the caching logic could be used elsewhere too? If so, should we extract it and move it to message for example?

src/Generator/RequestLineAndBodyGenerator.php
Nyholm: Yes please! Any suggestions?

Nyholm:
Thank you for the feedback. I've updated. I renamed the generator to "SimpleGenerator" since it uses an naive approach. 

>My fear is that we are putting too much logic here. Do you think the caching logic could be used elsewhere too? If so, should we extract it and move it to message for example?

The `CachePlugin` is pretty fat, yes. One could refactor it to more helper classes and strategies. 

Im not sure it make sense to try to reuse the cache logic. Not at this point, because the code it too coupled to this specific use case. But I will not close that door forever. 

If I recall correctly we decided to not have cache logic in message because we did not want to have so many external dependencies in php-http/message. But you may be referring to the HTTP specific parts... Again, not sure.

I think we can let this library grow and mature. It is an implementation (and not an abstraction like most of our other libraries) so it is fine to bump it to 2.x and 3.x when we see a need for it.

dbu (approved):
i like this and think its fine to put in here at least for now. lets improve the cache plugin and if we realize we need the cache key elsewhere we can always push it upstream in a BC way as the cache depends on client-common

Nyholm:
Thank you for the reviews.

dbu:
please add documentation for this. i created https://github.com/php-http/documentation/issues/187 so we don't forget.

Nyholm:
Thank you

Issue #31, opened 2017-03-21
different cache key strategy for client and server mode


dbu:
once #30 is done, we should implement different cache key strategies for the client and server mode of the plugin. as client, the key should take into account authorization headers (Authenticate, Cookies). in server mode, requests with cookies and or authentication should not be cached at all.

PR #36, opened 2017-03-31
Added VaryGenerator

### LINKED ISSUE #30: Customize cache key generation

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #30
| Documentation   | 
| License         | MIT

#### What's in this PR?

This is an other generator. This PR is blocked by #35

#### Why?

This generator will be used by KNP Github API. Related to https://github.com/KnpLabs/php-github-api/issues/557

### DIFF
--- a/src/Cache/Generator/HeaderCacheKeyGenerator.php
+++ b/src/Cache/Generator/HeaderCacheKeyGenerator.php
@@ -5,7 +5,7 @@ namespace Http\Client\Common\Plugin\Cache\Generator;
 use Psr\Http\Message\RequestInterface;
 
 /**
- * Generate a cache key by specify what headers you want to vary on.
+ * Generate a cache key by using HTTP headers.
  *
  * @author Tobias Nyholm <tobias.nyholm@gmail.com>
  */
--- a/src/Cache/Generator/HeaderCacheKeyGenerator.php
+++ b/src/Cache/Generator/HeaderCacheKeyGenerator.php
@@ -33,6 +33,6 @@ class HeaderCacheKeyGenerator implements CacheKeyGenerator
             $concatenatedHeaders[] = sprintf(' %s:"%s"', $headerName, $request->getHeaderLine($headerName));
         }
 
-        return $request->getMethod().' '.$request->getUri().implode('', $concatenatedHeaders);
+        return $request->getMethod().' '.$request->getUri().implode('', $concatenatedHeaders).' '.$request->getBody();
     }
 }
--- a/src/Cache/Generator/HeaderCacheKeyGenerator.php
+++ b/src/Cache/Generator/HeaderCacheKeyGenerator.php
@@ -9,7 +9,7 @@ use Psr\Http\Message\RequestInterface;
  *
  * @author Tobias Nyholm <tobias.nyholm@gmail.com>
  */
-class HeaderHashGenerator implements CacheKeyGenerator
+class HeaderCacheKeyGenerator implements CacheKeyGenerator
 {
     /**
      * The header names we should take into account when creating the cache key.
--- a/src/Generator/VaryGenerator.php
+++ b/src/Generator/VaryGenerator.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace Http\Client\Common\Plugin\Generator;
+
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * Generate a cache key and specify what headers you want to vary on.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class VaryGenerator implements CacheKeyGenerator
+{
+    /**
+     * The header names we should vary on.
+     *
+     * @var array
+     */
+    private $headerNames;
+
+    /**
+     * @param $headerNames
+     */
+    public function __construct(array $headerNames)
+    {
+        $this->headerNames = $headerNames;
+    }
+
+    public function generate(RequestInterface $request)
+    {
+        $concatenatedHeaders = [];
+        foreach ($this->headerNames as $headerName) {
+            $concatenatedHeaders[] = sprintf(' %s:"%s"', $headerName, $request->getHeaderLine($headerName));
+        }
+
+        return $request->getMethod().' '.$request->getUri().implode('', $concatenatedHeaders);
+    }
+}
--- a/spec/Cache/Generator/HeaderCacheKeyGeneratorSpec.php
+++ b/spec/Cache/Generator/HeaderCacheKeyGeneratorSpec.php
@@ -5,7 +5,7 @@ namespace spec\Http\Client\Common\Plugin\Cache\Generator;
 use PhpSpec\ObjectBehavior;
 use Psr\Http\Message\RequestInterface;
 
-class HeaderHashGeneratorSpec extends ObjectBehavior
+class HeaderCacheKeyGeneratorSpec extends ObjectBehavior
 {
     public function let()
     {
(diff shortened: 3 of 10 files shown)


src/Generator/VaryGenerator.php
```diff
@@ -0,0 +1,38 @@
...
+namespace Http\Client\Common\Plugin\Generator;
+
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * Generate a cache key and specify what headers you want to vary on.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class VaryGenerator implements CacheKeyGenerator
```
sagikazarmark: Should probably be final

Nyholm: Thank you for the review. 
I cannot make it final because PHP spec need to extend this class to be able to run the tests.

Nyholm:
Reading @dbu's ticket #31, this might be the wrong approach. Or at least redundant. One could argue that we should use `ClientKeyGenerator` instead.

dbu:
not sure about this one. the "proper" use case would be to work around a broken server that is not sending the correct vary header for its content. as i understand, you want it to use the github api which likely sends correct headers.

i think what i don't like about VaryGenerator that this is suggesting a relation with the Vary response header, which is not the case. could we call it HeaderHashGenerator? building the hash on Cookie and Authorization sounds ok. the proxy mode of the cache plugin could then use this generator by default.

> that we should use ClientKeyGenerator instead.

i think for the client mode, we don't need to differentiate on credentials, as its assuming we send requests for the same client, and we cache things with the private cache-control. the problem comes when you are in a multiple users mode, where the http client acts as some sort of proxy between the user and the service that is cached.

Nyholm:
Thank you. I've renamed the client. 

> i think for the client mode, we don't need to differentiate on credentials

Don't we? How would we solve the Gitbub issue? 

```php
// Request 1
$gitbub->setAuthenticationToken('foo');
$response = $github->me(); // MISS
// This will be stored in PSR6 cache

// Request 2, 2 days later
$gitbub->setAuthenticationToken('bar');
$response = $github->me(); // HIT
// This will fetch the cached response. 
// Github servers will never see this request. 
```

Should this be considered "Client mode" or "Server mode"?

dbu:
to me, that would be server mode. but i realize that the naming could be confusing... my idea was that client mode would set up things like when you use a web browser, you are one single user and don't change authentication for example.

do you have an idea for better names than client and server? single and shared maybe?

Nyholm:
You are correct. This is server mode. We need some good docs for this. 

Do we want to merge a generator like this now, or do we just one the ServerCacheGenerator and ClientCacheGenerator?

dbu:
i actually quite like my idea to move from the words client and server to single and shared. it might be more intuitive what this is about, or at least avoid assumptions that people get when we talk about server and client. then this could be a SharedCacheKeyGenerator and default to differentiate on cookie and authentication header. wdyt?

Nyholm:
I like it. I'll make some updates now

src/Cache/Generator/HeaderCacheKeyGenerator.php
```diff
@@ -0,0 +1,38 @@
...
+    public function __construct(array $headerNames)
+    {
+        $this->headerNames = $headerNames;
+    }
+
+    public function generate(RequestInterface $request)
+    {
+        $concatenatedHeaders = [];
+        foreach ($this->headerNames as $headerName) {
+            $concatenatedHeaders[] = sprintf(' %s:"%s"', $headerName, $request->getHeaderLine($headerName));
```
dbu, line 33: calling getHeaderLine for a header that is not defined simply returns an empty string. just double-checked in the psr-7 spec. so this is fine

src/Cache/Generator/HeaderCacheKeyGenerator.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+namespace Http\Client\Common\Plugin\Cache\Generator;
+
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * Generate a cache key by specify what headers you want to vary on.
```
dbu: i'd like to use a different word than "vary" here to avoid confusion with the `Vary` response header.

dbu (commented):
just a nitpick on the phpdoc, otherwise as per our slack discussion i think this is useful to have.

src/Cache/Generator/HeaderCacheKeyGenerator.php
Nyholm: :+1:

src/Cache/Generator/HeaderCacheKeyGenerator.php
Nyholm: Thank you for the review

src/Cache/Generator/HeaderCacheKeyGenerator.php
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    public function generate(RequestInterface $request)
+    {
+        $concatenatedHeaders = [];
+        foreach ($this->headerNames as $headerName) {
+            $concatenatedHeaders[] = sprintf(' %s:"%s"', $headerName, $request->getHeaderLine($headerName));
+        }
+
+        return $request->getMethod().' '.$request->getUri().implode('', $concatenatedHeaders);
```
dbu: to be consistent with the simple generator, we should also add the body. we might just extend the simple generator and concat parent::generate with the headers

dbu (commented):
sorry, another thing. noticed that this generator is not constistent with the simple generator in not using the body.

src/Cache/Generator/HeaderCacheKeyGenerator.php
Nyholm: I believe in cohesion over inheritance here. We should create some kind of chaining or use a separate class for chaining

dbu: we can also just add the body from simple generator. its just one line of code, i think we don't need to get too complicated with this. but we should have the body to not have inconsistent behaviour between the two generators

Nyholm: Okey, done. 

We do not need the if(empty()) check because it was added in the SimpleGenerator because of BC.

dbu:
great, thanks a lot! can you update the doc for this as well please? and should we start using this in the plugin when in server mode?

Nyholm:
> great, thanks a lot! can you update the doc for this as well please? 

Yes!

> and should we start using this in the plugin when in server mode?

That would be a BC break, right?

dbu:
> That would be a BC break, right?

i would consider it a bugfix. it will lead to cache misses after a
deployment, but i think its worth it because it avoids major accidents.

dbu:
actually as i want to rename the methods to single and shared anways
(for better understanding what the two modes are about), we could add
those and just leave client and server untouched and deprecate them.

Nyholm:
👍

PR #38, opened 2017-04-05
Only throw deprecation when user explicitly sets option

### LINKED ISSUE #37: A deprecation notice is always triggered
It is for `respect_cache_headers`. 

From @GrahamCampbell https://github.com/KnpLabs/php-github-api/issues/557#issuecomment-291838413

acrobat:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #37 
| Documentation   | /
| License         | MIT

#### What's in this PR?

This pr fixes the issue so the deprecation notice is only throw when the users sets the "respect_cache_headers" option.

### DIFF
--- a/src/CachePlugin.php
+++ b/src/CachePlugin.php
@@ -338,7 +338,7 @@ final class CachePlugin implements Plugin
             'cache_lifetime' => 86400 * 30, // 30 days
             'default_ttl' => 0,
             //Deprecated as of v1.3, to be removed in v2.0. Use respect_response_cache_directives instead
-            'respect_cache_headers' => true,
+            'respect_cache_headers' => null,
             'hash_algo' => 'sha1',
             'methods' => ['GET', 'HEAD'],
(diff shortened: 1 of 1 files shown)


Nyholm (approved):
Thank you

src/CachePlugin.php
```diff
@@ -338,7 +338,7 @@ private function configureOptions(OptionsResolver $resolver)
             'cache_lifetime' => 86400 * 30, // 30 days
             'default_ttl' => 0,
             //Deprecated as of v1.3, to be removed in v2.0. Use respect_response_cache_directives instead
-            'respect_cache_headers' => true,
+            'respect_cache_headers' => null,
```
Nyholm, line 341: You do also have to add "null" as an allowed type.

acrobat: Yes indeed, fixed!

Nyholm:
Thank you

PR #43, opened 2017-07-29
The etag from data can never be a string

GrahamCampbell:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | maybe?
| Deprecations?   | no
| License         | MIT

I don't think this key can ever produce anything other than a string, since it is set by `$response->getHeader('ETag')` which is always of type `string[]`.

Is this there for historical reasons perhaps? If so, instead of this PR, a comment should be added, similar to to the isset removal comment.

### DIFF
--- a/src/CachePlugin.php
+++ b/src/CachePlugin.php
@@ -435,10 +435,6 @@ final class CachePlugin implements Plugin
             return;
         }
 
-        if (!is_array($data['etag'])) {
-            return $data['etag'];
-        }
-
         foreach ($data['etag'] as $etag) {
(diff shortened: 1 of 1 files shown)


dbu:
@Nyholm added this with the introduction of e-tag support in 4bf191d7c984d960f9383a9381c6f6cbdb2e011e - not sure if there is something that could mess this up?

Nyholm (approved):
You are correct. This if statement will never evaluate to true

PR #48, opened 2018-01-15
Adding response mutator.

iainconnor:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #X, partially #Y, mentioned in #Z
| Documentation   | https://github.com/php-http/documentation/pull/221
| License         | MIT

#### What's in this PR?

Adds an optional mutator callback so you can mutate the response depending whether it was from cache or not.

#### Why?

Right now, there's no way to tell whether a response came from the cache or not, other than reverse-engineering the cache logic. It may be useful for debugging or tweaking your cache to know this, so this PR provides an optional way of retrieving this information. An `AddHeaderResponseMutator` is provided as an example way of adding a custom header to the response identifying whether it came from cache or not.

#### Example Usage

``` php
CachePlugin::clientCache($cacheItemPool, $streamFactory, ['response_mutator' => new AddHeaderResponseMutator('X-From-Cache')]);
```

#### Checklist

- [x] Updated CHANGELOG.md to describe BC breaks / deprecations | new feature | bugfix
- [x] Documentation pull request created (if not simply a bugfix)

#### To Do

- [x] Not sure how to update CHANGELOG.md. Should I be tagging a new version?


src/CachePlugin.php
```diff
@@ -357,6 +367,7 @@ private function configureOptions(OptionsResolver $resolver)
 
             return empty($matches);
         });
+        $resolver->setAllowedTypes('response_mutator', ['null', 'Http\Client\Common\Plugin\Cache\Mutator\ResponseMutator']);
```
Nyholm: We could use the `::class` constant

src/Mutator/AddHeaderResponseMutator.php
```diff
@@ -0,0 +1,37 @@
...
+ */
+class AddHeaderResponseMutator implements ResponseMutator
+{
+    /** @var string */
+    private $headerName;
+
+    /**
+     * @param string $headerName
+     */
+    public function __construct($headerName)
```
Nyholm: How about adding a default value here? I guess most people will be happy with any default.

Nyholm (approved):
This question has came up before. I do not really see the value of it. But this is a good implementation and I do not see why we should not allow it.

src/Mutator/AddHeaderResponseMutator.php
iainconnor: Agreed, I'll update with a sane default.

src/CachePlugin.php
iainconnor: I agree, but I was following the style of the other lines in this method.

src/CachePlugin.php
```diff
@@ -441,4 +452,22 @@ private function getETag(CacheItemInterface $cacheItem)
...
+     * @param RequestInterface  $request
+     * @param ResponseInterface $response
+     * @param bool              $cacheHit
+     *
+     * @return ResponseInterface
+     */
+    private function handleResponseMutator(RequestInterface $request, ResponseInterface $response, $cacheHit)
+    {
+        if (null !== $this->config['response_mutator']) {
+            $response = $this->config['response_mutator']->mutate($request, $response, $cacheHit);
```
dbu: i think this should be a list of mutators.

src/CachePlugin.php
dbu: lets use it here and we fix it elsewhere.

(but as saying below, i think this should be an array, while we are at it.

src/Cache/Mutator/AddHeaderResponseMutator.php
```diff
@@ -0,0 +1,38 @@
...
+ */
+class AddHeaderResponseMutator implements ResponseMutator
+{
+    /** @var string */
+    private $headerName;
+
+    /**
+     * @param string $headerName
+     */
+    public function __construct($headerName = 'X-From-Php-Http-Cache')
```
dbu: a semi-official way to mark cache hits is to use "X-Cache: HIT" (resp. MISS). can we find a different example?

src/Cache/Mutator/AddHeaderResponseMutator.php
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Mutate the response depending on the cache status.
+     *
+     * @param ResponseInterface $response
+     * @param bool              $cacheHit
+     *
+     * @return string
+     */
+    public function mutate(RequestInterface $request, ResponseInterface $response, $cacheHit)
```
dbu: is there any more information that we could pass in instead of the boolean? like the $cacheItem or such? then this could be used to extract additional debugging information if wanted.

src/Cache/Mutator/ResponseMutator.php
```diff
@@ -0,0 +1,25 @@
...
+
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+
+/**
+ * An interface for a mutator of a possibly cached response.
+ *
+ * @author Iain Connor <iain.connor@priceline.com>
+ */
+interface ResponseMutator
```
dbu: i am not happy with this name. every plugin has the possibility to change a response. this thing is specifically about the cache. but unfortunately i don't have a great name either. `CacheListener` is not too good, but would be more specific to what this does. can we be even more specific?

the class doc could be "XXX is called by the cache plugin with information on the cache status to change information on the response based on whether it was a cache hit or a miss."

dbu (commented):
thanks for this one! the architecture looks clean to me.

a way to achieve this with existing code features would be to put a plugin into the chain after the cache, and one before the cache. the after adds the MISS value, the one before checks if X-Cache is set and if not, its a HIT. but i agree that that would be quite an incidental way of doing things.

are there any uses cases other than adding an X-Cache: HIT/MISS header? i think either we should find reasons why additional flexibility is needed, or simply have a flag (or header name config) to make the cache plugin itself add the hit or miss header. other plugins could then look at that header to know if they need to further change the response, if you need to change a response based on whether it comes from the cache. if such a scenario exists at all... maybe its better to just have the header, and if somebody has a really special need, they can always implement their own cache plugin based on this repo?

iainconnor:
> are there any uses cases other than adding an X-Cache: HIT/MISS header?

The other use-case that my project is currently adding is logging of the hit/miss rate. I realize I could probably accomplish that in another way (by adding the header and then checking for the presence of that header, for example), but adding it the same plugin seemed clearer.

src/Cache/Mutator/AddHeaderResponseMutator.php
iainconnor: Good suggestion, updated.

src/Cache/Mutator/AddHeaderResponseMutator.php
iainconnor: Good suggestion, updated.

iainconnor: For the record, I'm going to keep the `cacheHit` parameter as well, so the consumer doesn't have to reverse-engineer the logic regarding expiry. Checking `cacheItem->isHit()` won't be sufficient.

src/Cache/Mutator/ResponseMutator.php
iainconnor: I can't think of anything significantly better. Will rename this for the time being and see how it feels.

src/CachePlugin.php
iainconnor: Good suggestion, will update.

iainconnor: Though, not totally familliar with the resolver... is there any better way of stating that the array must be of generic type `CacheListener`? Right now I just have `$resolver->setAllowedTypes('cache_listeners', ['array']);`.

dbu (approved):

dbu:
@Nyholm want another look?

also, should we support this in the symfony bundle too?

iainconnor:
Saw the documentation get merged in. Anything else needed for this code?

dbu:
ups, somehow missed to merge this. the changelog is still open. can you please update composer.json to make the branch alias 1.6 and add a changelog entry for 1.6.0 that says this was added (short, but say what its for)?

iainconnor:
Sure thing; done.

Nyholm (approved):

dbu:
yay, thanks a lot!

PR #56, opened 2019-06-05
Support PSR-17 StreamFactoryInterface

deguif:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no? 
| Deprecations?   | no
| Related tickets | https://github.com/php-http/HttplugBundle/pull/339
| Documentation   | 
| License         | MIT

#### What's in this PR?

Support for PSR-17 `StreamFactoryInterface`

### DIFF
--- a/src/CachePlugin.php
+++ b/src/CachePlugin.php
@@ -13,6 +13,7 @@ use Psr\Cache\CacheItemInterface;
 use Psr\Cache\CacheItemPoolInterface;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\StreamFactoryInterface;
 use Symfony\Component\OptionsResolver\Options;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 
@@ -33,7 +34,7 @@ final class CachePlugin implements Plugin
     private $pool;
 
     /**
-     * @var StreamFactory
+     * @var StreamFactory|StreamFactoryInterface
      */
     private $streamFactory;
 
@@ -50,9 +51,9 @@ final class CachePlugin implements Plugin
     private $noCacheFlags = ['no-cache', 'private', 'no-store'];
 
     /**
-     * @param CacheItemPoolInterface $pool
-     * @param StreamFactory          $streamFactory
-     * @param array                  $config        {
+     * @param CacheItemPoolInterface               $pool
+     * @param StreamFactory|StreamFactoryInterface $streamFactory
+     * @param array                                $config        {
      *
      *     @var bool $respect_cache_headers Whether to look at the cache directives or ignore them
      *     @var int $default_ttl (seconds) If we do not respect cache headers or can't calculate a good ttl, use this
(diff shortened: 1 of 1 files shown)


deguif:
Build is failing because PHP 5.4 is still supported (unsupported `::class` constant), maybe support for unmaintained PHP versions could be dropped?

shulard:
I think you are right. The whole PHP-HTTP ecosystem is moving around the v2 so its time to deprecate some old PHP versions. Regarding that update, maybe supporting both HttplugFactory and PSR17Factory is not relevant anymore... PSR17 will be the de facto standard.

dbu:
not just "de facto" but "officially accepted" standard :-)

afaik @Nyholm did a lot of thinking to have a clean upgrade path for the PSR, maybe he can chime in on this?

deguif:
Yes, let's wait for his comment(s). 
I would be very interested to know about the upgrade path ;)

dbu:
@Nyholm do you have input on this one?

dbu:
ping @Nyholm

Nyholm (approved):
Im sorry for the awful delay. 

This looks super solid.

src/CachePlugin.php
```diff
@@ -66,8 +67,12 @@ final class CachePlugin implements Plugin
      *              Defaults to an empty array
      * }
      */
-    public function __construct(CacheItemPoolInterface $pool, StreamFactory $streamFactory, array $config = [])
+    public function __construct(CacheItemPoolInterface $pool, $streamFactory, array $config = [])
     {
+        if (!($streamFactory instanceof StreamFactory) && !($streamFactory instanceof StreamFactoryInterface)) {
+            throw new \LogicException(\sprintf('StreamFactory must be an instance of %s or %s', StreamFactory::class, StreamFactoryInterface::class));
```
GrahamCampbell: Why don't we throw a `TypeError` here?

GrahamCampbell (changes requested):

src/CachePlugin.php
dbu: oh good idea. @deguif could you change this to throw a TypeError instead?

deguif: Yes I can change this ;)

deguif: And ✅

src/CachePlugin.php
```diff
@@ -69,8 +70,12 @@ final class CachePlugin implements Plugin
      *              Defaults to an empty array
      * }
      */
-    public function __construct(CacheItemPoolInterface $pool, StreamFactory $streamFactory, array $config = [])
+    public function __construct(CacheItemPoolInterface $pool, $streamFactory, array $config = [])
     {
+        if (!($streamFactory instanceof StreamFactory) && !($streamFactory instanceof StreamFactoryInterface)) {
+            throw new \TypeError(\sprintf('StreamFactory must be an instance of %s or %s.', StreamFactory::class, StreamFactoryInterface::class));
```
GrahamCampbell: Should be:

```
throw new \TypeError(\sprintf('Argument 2 passed to %s::__construct() must be of type %s|%s, %s given.', self::class, StreamFactory::class, StreamFactoryInterface::class, gettype($streamFactory)));
```

To match the message given by PHP 8 for type errors with union types.

GrahamCampbell (changes requested):

src/CachePlugin.php
deguif: Even better: `\is_object($streamFactory) ? \get_class($streamFactory) : \gettype($streamFactory)` for the given type ;)

dbu (approved):
thanks a lot!

src/CachePlugin.php
```diff
@@ -33,7 +34,7 @@ final class CachePlugin implements Plugin
     private $pool;
 
     /**
-     * @var StreamFactory
+     * @var StreamFactory|StreamFactoryInterface
```
dbu, line 37: once https://github.com/php-http/message-factory/pull/38 is released, we can raise our version constraint to get that version and only use the psr interface from then on.

deguif: Yes, and let ping me at this moment, I will be glad to do the PR ;)

src/CachePlugin.php
GrahamCampbell: Ah, yes. 👍

PR #58, opened 2019-06-05
Clean Symfony versions

deguif:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| Documentation   |
| License         | MIT

#### What's in this PR?

Remove unmaintained Symfony versions.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
         "psr/cache": "^1.0",
         "php-http/client-common": "^1.9 || ^2.0",
         "php-http/message-factory": "^1.0",
-        "symfony/options-resolver": "^2.6 || ^3.0 || ^4.0"
+        "symfony/options-resolver": "^3.4.20 || ^4.2.1"
     },
     "require-dev": {
         "phpspec/phpspec": "^2.5",


deguif:
Fully inspired by https://github.com/php-http/HttplugBundle/pull/343

GrahamCampbell:
This seems too brutal tbh. Like, all that will happen here that composer will start to resolve older patch releases of this package in order to let people use older symfony. Is this really what we want?

deguif:
You're right about the impact.
This will have more impact here than in the HttplugBundle. Probably a bad idea ;)

dbu:
my position is: if we get hindered by something (new features, BC breaks etc) we should not hesitate to restrict to only supported symfony versions. but without the need for it, i see no harm in allowing them.

the options resolver probably did not change much over all these versions, so we don't gain anything by excluding older versions.

deguif:
Yes, I should have thought twice before opening this PR. Let's close it.

dbu:
:+1:  and no worries, i rather have merge requests that we end up not merging than people not caring ;-)

if you want to contribute something, how about picking up https://github.com/php-http/HttplugBundle/pull/319 ? i won't get to work on that this month, so its up for grabs. or there is a couple of issues for the cache-plugin that could be interesting to implement (and require less symfony configuration knowhow) https://github.com/php-http/cache-plugin/issues

PR #61, opened 2019-11-25
Added blacklisted_paths option

schroedingerskatze42:
- adds a list of strings (regular expressions) to describe patterns of paths not to be cached.

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    |yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| Documentation   | -
| License         | MIT

#### What's in this PR?

Added an option 'blacklisted_paths'; a list of regular expressions to describe paths not to be cached. 

#### Why?

This is necessary if a third party resource is requested of providers, who don't get the usage of response cache headers. (But will return no-cache every time, for every endpoint)  

#### Example Usage

``` php
$cachePlugin = new CachePlugin($pool, $streamFactory, ['blacklisted_paths' => [
  '^\/this-path-shall\/not-been-cached\/.*',
]]);
```

#### Checklist

- [X] Updated CHANGELOG.md to describe BC breaks / deprecations | new feature | bugfix
- [X] Documentation pull request created: https://github.com/php-http/documentation/pull/274

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Change Log
 
-## 1.6.1 - 2019-11-25
+## 1.7.0 - unreleased
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## 1.6.1 - 2019-11-25
+
+### Added
+
+* Added `blacklisted_paths` option, which takes an array of `strings` (regular expressions) and allows to define paths, that shall not be cached in any case.
+
 ## 1.6.0 - 2019-01-23
 
 ### Added
--- a/src/CachePlugin.php
+++ b/src/CachePlugin.php
@@ -274,7 +274,7 @@ final class CachePlugin implements Plugin
     protected function isCacheableRequest(RequestInterface $request)
     {
         foreach ($this->config['blacklisted_paths'] as $not_to_cache_path) {
-            if (1 === preg_match('/'.$not_to_cache_path.'/', $request->getRequestTarget())) {
+            if (1 === preg_match('/'.$not_to_cache_path.'/', $request->getUri())) {
                 return false;
             }
         }
--- a/src/CachePlugin.php
+++ b/src/CachePlugin.php
@@ -76,7 +76,6 @@ final class CachePlugin implements Plugin
             throw new \InvalidArgumentException('You can\'t provide config option "respect_cache_headers" and "respect_response_cache_directives". '.'Use "respect_response_cache_directives" instead.');
         }
 
-
         $optionsResolver = new OptionsResolver();
         $this->configureOptions($optionsResolver);
         $this->config = $optionsResolver->resolve($config);
--- a/src/CachePlugin.php
+++ b/src/CachePlugin.php
@@ -247,7 +247,7 @@ final class CachePlugin implements Plugin
     /**
      * Verify that we can cache this response.
      *
-     * @param RequestInterface $request
+     * @param RequestInterface  $request
      * @param ResponseInterface $response
      *
      * @return bool
--- a/src/CachePlugin.php
+++ b/src/CachePlugin.php
@@ -267,7 +267,7 @@ final class CachePlugin implements Plugin
     /**
      * Verify that we can cache this request.
      *
-     * @param RequestInterface  $request
+     * @param RequestInterface $request
      *
      * @return bool
      */
--- a/src/CachePlugin.php
+++ b/src/CachePlugin.php
@@ -181,7 +181,7 @@ final class CachePlugin implements Plugin
                 return $this->handleCacheListeners($request, $this->createResponseFromCacheItem($cacheItem), true, $cacheItem);
             }
 
-            if ($this->isCacheable($request, $response)) {
+            if ($this->isCacheable($response) && $this->isCacheableRequest($request)) {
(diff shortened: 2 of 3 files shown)


src/CachePlugin.php
```diff
@@ -72,10 +73,7 @@ public function __construct(CacheItemPoolInterface $pool, StreamFactory $streamF
         $this->streamFactory = $streamFactory;
 
         if (isset($config['respect_cache_headers']) && isset($config['respect_response_cache_directives'])) {
-            throw new \InvalidArgumentException(
-                'You can\'t provide config option "respect_cache_headers" and "respect_response_cache_directives". '.
-                'Use "respect_response_cache_directives" instead.'
-            );
+            throw new \InvalidArgumentException('You can\'t provide config option "respect_cache_headers" and "respect_response_cache_directives". '.'Use "respect_response_cache_directives" instead.');
```
dbu: did the codestyle checker complain about this before? otherwise lets not touch it.

if codestyle complains, lets remove the concatenation and make it all one string.

src/CachePlugin.php
```diff
@@ -353,13 +358,15 @@ private function configureOptions(OptionsResolver $resolver)
             'respect_response_cache_directives' => ['no-cache', 'private', 'max-age', 'no-store'],
             'cache_key_generator' => null,
             'cache_listeners' => [],
+            'blacklisted_paths' => [],  // restricted for
```
dbu: please remove the comment, the phpdoc on the constructor should document this (plus the documentation change)

src/CachePlugin.php
```diff
@@ -246,16 +244,23 @@ private function calculateResponseExpiresAt($maxAge)
     /**
      * Verify that we can cache this response.
      *
+     * @param RequestInterface  $request
      * @param ResponseInterface $response
      *
      * @return bool
      */
-    protected function isCacheable(ResponseInterface $response)
+    protected function isCacheable(RequestInterface $request, ResponseInterface $response)
```
dbu: i wonder why this method is protected. extending the plugin sounds like a bad idea, but maybe that was the intention. as it is protected, this is technically a BC break. we could separate this into another method `isCacheableRequest` to avoid the BC break.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Change Log
 
+## 1.6.1 - 2019-11-25
```
dbu: this is a new feature and should lead to 1.7.0

please replace the date with "unreleased"

dbu (commented):
thanks for contributing. this is rather an edge case, but as we do have modes of ignoring caching instructions from the server, it seems reasonable to me to have this config.

could you please add something to the test in spec/CachePluginSpec.php to have one example of blacklisting tested? and wdyt of splitting it into a separate method to avoid the BC break?

src/CachePlugin.php
schroedingerskatze42: yes, the code style checker complained about it; should already have been existed before. no magic button (auto format) used.

regarding the test: WIP, give me 15 minutes.

CHANGELOG.md
schroedingerskatze42: done; also replaced 1.6.1 with 1.7.0; not quite sure, what is right here. feel free to enlighten me :-)

src/CachePlugin.php
schroedingerskatze42: you are right. i outsourced the checking of the request into the method isCacheableRequest and reset the "isCacheable" to it's original state.

The only thing, I am bothering about is the naming (one of the two hard things of computer science). If we add a method "isCacheableRequest", the method "isCacheable" should be renamed to "isCacheableResponse" and a new method "isCacheable" should take both arguments and return the result of isCacheableRequest && isCachableResponse().

This however would lead again to the BC, we would like to avoid.

src/CachePlugin.php
schroedingerskatze42: done

src/CachePlugin.php
dbu: the stupid thing is that it does a string concat on that one line now, for no good reason... can you remove the `'.'` in the middle please?

src/CachePlugin.php
dbu: yep. my idea would be that in version 2 we make the plugin final, this method private and can change the signature.

CHANGELOG.md
dbu: that looks good, thanks.

src/CachePlugin.php
```diff
@@ -266,6 +264,24 @@ protected function isCacheable(ResponseInterface $response)
...
     }
 
+    /**
+     * Verify that we can cache this request.
+     *
+     * @param RequestInterface  $request
+     *
+     * @return bool
+     */
+    protected function isCacheableRequest(RequestInterface $request)
```
dbu: please make this method private.

dbu:
apart from my nitpicks, i am happy with this.

@nobYsDarling can you please do the whitespace fix that codestyle is complaining about?

@Nyholm any inputs from your side or happy with this?

src/CachePlugin.php
schroedingerskatze42: done.

src/CachePlugin.php
schroedingerskatze42: done.

dbu (approved):

PR #64, opened 2019-12-16
Prepare for 1.7.0

Nyholm:
Im happy to release 1.7.0 after #56 is merged


dbu:
i'd love to also clarify #62 before we release - we did not yet release a version with that new feature, so adjustions now won't break stable applications.

Nyholm:
#62 is now clarified in #65.

PR #65, opened 2019-12-26
Bugfixes with blacklist

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no (has not been released yet)
| Deprecations?   | no
| Related tickets | 
| Documentation   | 
| License         | MIT

This will fix #62 

The BC breaking change is that we allow any regex delimiter.

### DIFF
--- a/src/CachePlugin.php
+++ b/src/CachePlugin.php
@@ -63,7 +63,7 @@ final class CachePlugin implements Plugin
      *              we have to store the cache for a longer time than the server originally says it is valid for.
      *              We store a cache item for $cache_lifetime + max age of the response.
      *     @var array $methods list of request methods which can be cached
-     *     @var array $blacklisted_paths list of regex of to URLs explicitly not to be cached
+     *     @var array $blacklisted_paths list of regex for URLs explicitly not to be cached
      *     @var array $respect_response_cache_directives list of cache directives this plugin will respect while caching responses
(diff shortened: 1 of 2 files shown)


src/CachePlugin.php
```diff
@@ -63,7 +63,7 @@ final class CachePlugin implements Plugin
      *              we have to store the cache for a longer time than the server originally says it is valid for.
      *              We store a cache item for $cache_lifetime + max age of the response.
      *     @var array $methods list of request methods which can be cached
-     *     @var array $blacklisted_paths list of regex patterns of paths explicitly not to be cached
+     *     @var array $blacklisted_paths list of regex of to URLs explicitly not to be cached
```
gmponos: Shouldn't it be just `of URLs`

gmponos (approved):
A small typo change only.

Although it is not related with the current PR please consider at some point refactoring the tests and using actual objects.. For instance if you used `ArrayAdapter` and an actual `Request` object I believe some of the tests changes there would not be needed now.

Anyway as I said it is not blocking for the current PR.

src/CachePlugin.php
Nyholm: Oh,, I'll go with "for"

Nyholm:
Yeah, But that is not how PHPSpec is working (as far as I know). But I would love to do that. =)

Thank you for the review

PR #74, opened 2021-10-25
add phpstan to ci build

### LINKED ISSUE #73: Warning: array_key_exists() expects parameter 2 to be array, null given
**PHP version**: 7.3

**Description**

Getting these warnings reported in Sentry on some requests. The chance of this error happening is roughly 1%.

image

https://github.com/php-http/cache-plugin/blob/master/src/CachePlugin.php#L150

**How to reproduce**
Not entirely sure to be honest. Using the PSR16 adapter of https://www.phpfastcache.com/ with the Redis driver, and the Symfony cache adapter to convert PSR16 to PSR6. It only happens on a small subset of requests. I'm wondering if this is a race condition, when the cache item expires in between cache lookups. First call sees if item is in cache, by the time the second call tries to get the data, the item has expired.

**Possible Solution**

Check `$data` for `null` before processing further.

dbu:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #73
| Documentation   | -
| License         | MIT

#### What's in this PR?

Run phpstan on the codebase and fix things that phpstan discovers.

#### Why?

#73 reported that some cache implementations can have `null` data even if the cache is a hit.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -2,6 +2,8 @@ name: tests
 
 on:
   push:
+    branches:
+      - master
   pull_request:
 
 jobs:
--- a/src/CachePlugin.php
+++ b/src/CachePlugin.php
@@ -222,7 +222,7 @@ final class CachePlugin implements Plugin
     private function calculateCacheItemExpiresAfter($maxAge)
     {
         if (null === $this->config['cache_lifetime'] && null === $maxAge) {
-            return;
+            return null;
         }
 
         return $this->config['cache_lifetime'] + $maxAge;
@@ -239,7 +239,7 @@ final class CachePlugin implements Plugin
     private function calculateResponseExpiresAt($maxAge)
     {
         if (null === $maxAge) {
-            return;
+            return null;
         }
 
         return time() + $maxAge;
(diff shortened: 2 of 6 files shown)


.github/workflows/tests.yml
```diff
@@ -2,6 +2,8 @@ name: tests
 
 on:
   push:
+    branches:
+      - master
```
dbu, line 6: this prevents us from running every build twice

src/CachePlugin.php
```diff
@@ -146,22 +152,24 @@ protected function doHandleRequest(RequestInterface $request, callable $next, ca
...
             $data = $cacheItem->get();
-            // The array_key_exists() is to be removed in 2.0.
-            if (array_key_exists('expiresAt', $data) && (null === $data['expiresAt'] || time() < $data['expiresAt'])) {
-                // This item is still valid according to previous cache headers
-                $response = $this->createResponseFromCacheItem($cacheItem);
-                $response = $this->handleCacheListeners($request, $response, true, $cacheItem);
-
-                return new FulfilledPromise($response);
-            }
+            if (is_array($data)) {
```
dbu, line 155: phpstan does not complain, but CacheItem::get is defined to return `mixed`. in some situations, the cache might return something that is not an array.

PR #78, opened 2022-01-09
Added psr/cache:^3.0 dependency

yoshz:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | N/A
| Documentation   | N/A
| License         | MIT

#### What's in this PR?

Allow the usage of `psr/cache:3.0`.

#### Why?
Version `psr/cache:3.0` is already released a while ago and should be compatible with this plugin but the current composer.json configuration doesn't allow you to use this.
In my use case composer complains about a conflict while installing the cache-plugin because `symfony/cache:6.0` requires `psr/cache:3.0`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
     ],
     "require": {
         "php": "^7.1 || ^8.0",
-        "psr/cache": "^1.0 || ^2.0",
+        "psr/cache": "^1.0 || ^2.0 || ^3.0",
         "php-http/client-common": "^1.9 || ^2.0",
         "php-http/message-factory": "^1.0",
         "symfony/options-resolver": "^2.6 || ^3.0 || ^4.0 || ^5.0 || ^6.0"


GrahamCampbell:
I don't think this will work without changes to the code.

GrahamCampbell:
>In my use case composer complains about a conflict while installing the cache-plugin because symfony/cache:6.0 requires psr/cache:3.0.

Add this cache-plugin by hand to your composer.json and run `composer update`. It should install.

dbu:
thanks @yoshz for this. seems good to me, and the tests agree.

but lets give graham time to react:

> I don't think this will work without changes to the code.

https://github.com/php-http/cache-plugin/runs/4758238019?check_suite_focus=true installed psr/cache 3.0.0 and is green. either it works, or the tests do not trigger the problem.

looking at https://github.com/php-fig/cache/compare/2.0.0...3.0.0, i think the only difference is that the psr interfaces now have return types. we don't implement the interfaces in this package but only consume it, so it should be fine. what do you think shoudl be a problem, @GrahamCampbell

yoshz:
I also checked the changelog and the only difference is indeed the return types. So should be BC and I already tested it with my project.

Nyholm:
I was sure this would work. 

As a consumer of `psr/cache` you can depend on any version. The libraries that implement the interfaces are forced to do a new major version. 

Yes. I agree that this is just busy work that gives zero real benefit. But that is what the Fig decided to do. 

Im 👍🏽

dbu:
https://github.com/php-http/cache-plugin/releases/tag/1.7.5

Nyholm:
Thank you @yoshz and @dbu

PR #82, opened 2023-04-23
Detach stream before serializing PSR-7 response

### LINKED ISSUE #81: PSR-7 responses shouldn't be considered as serializable
In src/CachePlugin.php, PSR-7 responses are serialized as is, except the response body, which is serialized separately.

Recently the PHP engine started to throw deprecations when a resource is found in a serialized data structure.
And this happens to PSR-7 responses: whatever the implementation, their body is a represented as a stream resource internally.
Also, from an abstraction PoV, nothing says in PSR-7 that responses are serializable, which means interop should consider them as non-serializable.

The plugin should be improved to not store PSR-7 objects directly, but intermediate representations that fit a generic cache backend.

nicolas-grekas:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #81
| Documentation   | -
| License         | MIT

This might be enough to fix https://github.com/php-http/cache-plugin/issues/81

Feel free to amend at will :)

### DIFF
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -13,7 +13,7 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: PHPStan
         uses: docker://oskarstark/phpstan-ga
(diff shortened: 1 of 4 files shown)


dbu:
thanks a lot! i continue this in #84 as i rebased on 1.x to get the fixes for phpspec.

PR #86, opened 2023-05-30
Update composer.json

### LINKED ISSUE #85: php-http/message-factory is abandoned (deprecation error)
**PHP version**: x.y.z (hint: `php --version`)

all

**Description**

Package php-http/message-factory is abandoned, you should avoid using it. Use psr/http-factory instead.

**How to reproduce**

composer install with any pacakge that uses it.

**Possible Solution**

use psr/http-factory instead

tacman:
use the recommended package, fixes #85

| Q               | A
| --------------- | ---
| Deprecations?   | yes, fixes deprecation
| Related tickets | #85
| License         | MIT

#### What's in this PR?

Trivial change to library

#### Why?

#85

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         "php": "^7.1 || ^8.0",
         "psr/cache": "^1.0 || ^2.0 || ^3.0",
         "php-http/client-common": "^1.9 || ^2.0",
-        "php-http/message-factory": "^1.0",
+        "psr/http-factory": "^1.0",
         "symfony/options-resolver": "^2.6 || ^3.0 || ^4.0 || ^5.0 || ^6.0"
     },
     "require-dev": {


composer.json
```diff
@@ -14,7 +14,7 @@
         "php": "^7.1 || ^8.0",
         "psr/cache": "^1.0 || ^2.0 || ^3.0",
         "php-http/client-common": "^1.9 || ^2.0",
-        "php-http/message-factory": "^1.0",
+        "psr/http-factory": "^1.0",
```
dbu, line 17: we do not need just the factory interfaces but a factory implementation. the correct thing to require is [psr/http-factory-implementation](https://packagist.org/providers/psr/http-factory-implementation)

dbu (changes requested):
thanks for the pull request. we already support PSR-17 in the cache class.

the problem is that for bootstrapping, one has to pass an implementation of either the PSR-17 or the legacy php-http factory. 

i think to avoid hickups for the users, the best is to start a new major version of the cache plugin that requires a psr-17 implementation and no php-http/message-factory anymore and also does not allow the legacy anymore, so that we can declare types strictly again.

do you want to change the code accordingly? you will need to add an actual psr-17 implementation to require-dev and instantiate that in the tests. i suggest `nyholm/psr7`

composer.json
tacman: Hmm, while I follow what you're saying, I don't think I should be the one to change the code, as I don't understand it enough.

I'm just trying to get rid of deprecation messages, which are harder-than-expected to do in Symfony.

dbu: fair enough. i will see when i get around to do it.

this is only a deprecation warning, not an error, nothing should break and the legacy factories will stay available for a couple of years, we won't just delete the repository.

tacman: Yeah, this likely would be better as part of a 2.0 release -- where you could drop support for all EOL components, including PHP 7.  

Anyone needing the older code could keep using 1.7.  Then we can use type-hinting and method return types and all the other goodies that come with PHP 8.

Rector is a great tool for this.

danepowell:
@dbu @tacman is this something you plan to continue working on? I know the legacy library _probably_ won't disappear or suffer a vulnerability, but not everyone has the same risk tolerance. I for one will start getting pretty nervous soon 😄

tacman:
@danepowell realistically, I won't, as I don't understand it enough.  it's just one of the deprecation errors I'd like to get rid of, so I figured I'd report it here.

dbu:
the legacy library is done by php-http as well, as is the cache plugin. so it won't just disapear (or if we would somehow disapear the cache plugin would be gone as well - but i don't plan on disapearing ;-) ). there is no urgent risk here.

i don't have time right now to do the refactoring but would be happy to review if somebody has time to do it. i don't think it should be very complicated, as the interfaces are very similar and behave the same way.

dbu:
fixed in #87

PR #89, opened 2024-02-13
remove $ so gitclip works

tacman:
#### Why?

Otherwise, you get the $ on the command line and have to manually erase it.


dbu (approved):
thank you

PR #93, opened 2025-02-11
fix #91: Change clientCache so it doesn't force no-cache & max-age response directives

### LINKED ISSUE #91: 'respect_response_cache_directives' is ignored
We have the following code in CachePlugin.php:
```php
    public static function clientCache(CacheItemPoolInterface $pool, StreamFactoryInterface $streamFactory, array $config = [])
    {
        // Allow caching of private requests
        if (\array_key_exists('respect_response_cache_directives', $config)) {
            $config['respect_response_cache_directives'][] = 'no-cache';
            $config['respect_response_cache_directives'][] = 'max-age';
            $config['respect_response_cache_directives'] = array_unique($config['respect_response_cache_directives']);
        } else {
            $config['respect_response_cache_directives'] = ['no-cache', 'max-age'];
        }

        return new self($pool, $streamFactory, $config);
    }
```
If I pass a config with `'respect_response_cache_directives' => []`, it is ignored and replaced with `['no-cache', 'max-age']`.
I wanted it to always cache any page regardless of the server directives. It's impossible to do that right now. I'm not sure what's the intention of the code above.

nunoplopes:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | fixes #91
| License         | MIT

#### What's in this PR?

`clientCache()` will no longer force no-cache & max-age directives to be respected even when `respect_response_cache_directives` is set.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Change Log
 
+# Version 3
+
+## 3.0.0 - 2025-XX-XX
+
+### Changed
+ - `clientCache()` will no longer force no-cache & max-age directives to be respected even when `respect_response_cache_directives` is set.
+
 # Version 2
 
 ## 2.0.1 - 2024-10-02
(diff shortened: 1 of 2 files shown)


dbu:
nice, thanks!

